PR #46, opened 2016-03-03
A minor cleanup

jhrozek:
the patches in this PR get rid of some compilation warnings.


spbnick:
Thanks, Jakub! I have just one comment on one commit.

jhrozek:
I actually prefer the (void) approach :-) but I saw the attribute being used elsewhere so I thought that's what you prefer..

Let me resubmit the patches..

spbnick:
Thanks, Jakub!

PR #50, opened 2016-03-09
travis-ci integration

dkopecek:
Hello,
this is an attempt to integrate the Scribery/tlog repository with the travis-ci CI platform. I guess I'll need admin access to the organization to complete the integration, or you (@spbnick) will have to do it. What needs to be done is:
 - [x] Create the travis-ci Scribery account (log in via Github)
 - [x] Enable and configure the repository at travis-ci.org

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 *~
 .*
 !.git*
+!.travis.yml
 *.tar.gz
 *.o
 *.deps
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: c
+sudo: required
+dist: trusty
+compiler:
+- clang
+- gcc
+os: linux
+before_install:
+- sudo apt-get -qq update
+- sudo apt-get install -y libjson-c-dev libcurl4-gnutls-dev
+before_script:
+- autoreconf -i -f
+- ./configure --disable-silent-rules
+script:
+- make
+- make check
+- make distcheck
+
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 Tlog
 ====
 
+[![Build Status](https://travis-ci.org/Scribery/tlog.svg?branch=master)](https://travis-ci.org/Scribery/tlog)
+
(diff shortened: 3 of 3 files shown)


spbnick:
Thanks, Daniel! I made you an admin. Please set up the travis-ci side. I have two comments on the patch, though.

dkopecek:
@spbnick I use out-of-tree builds for mainly for 2 reasons:
 1. to not pollute the sources with various build generates files
 2. to catch bugs in Makefiles

Otherwise there are no real benefits I guess.

dkopecek:
@spbnick regarding --enable-debug, yes, we should do CI with that option as well. I'd like to add some testing matrix as we should test a combination of some options. I'll add that later, after we get to a working integration.

Another idea in the queue is code coverage of the test suite (another badge we can add to the main page). And of course the static code analysis. Both will come later via pull requests.

spbnick:
Well, I do use in-tree builds and I'd like to have that tested in CI, if possible. OTOH, `make distcheck` tests out-of-tree builds, AFAIK. Moreover it does that with the source tree read-only. Would that be enough to test your out-of-tree building? Or am I missing something?

Perhaps as `make distcheck` does a distribution first and then builds from a tarball instead of the actual git tree, it can miss some files which would interfere otherwise? Or would it be a problem for travis-ci if we did an in-tree build?

spbnick:
Otherwise, awesome ideas, I support them fully :+1: :)

dkopecek:
@spbnick ok, I'll change that to an in-tree build.

spbnick:
@dkopecek Thanks! Please feel free to raise the issue if it doesn't work for you after all.

spbnick:
@dkopecek Could you please squash these two commits into one?

dkopecek:
@spbnick done

spbnick:
Thanks!

spbnick:
Another idea can be building RPMs. Ubuntu should have rpmbuild, as Debian has it.

PR #51, opened 2016-03-10
[travis-ci] Run with several build configurations

dkopecek:
- plain configure build
 - test configure --enable-debug configure option
 - compile and run with code coverage instrumentation
   and send results to coveralls.io
 - run static code analysis using clang-analyzer


spbnick:
Looks great, thank you Daniel! Would you be able to resolve the issues found in the code and put the fixes before the integration patch into this pull request? If not, I can work on this, but later.

dkopecek:
@spbnick is there a reason for not using C99?

spbnick:
No, and I'm actually trying to enable it in configure.ac, but apparently
something goes wrong. It would be great, if you could fix that.

dkopecek:
@spbnick Ok, I'll look into the issue, but not as part of this PR.

dkopecek:
@spbnick it looks like that the issue is that clang's ccc-analyzer needs an explicit -std=c99 to accept c99 code whereas plain gcc/clang compilation does not. It is strange that the AC_PROG_STDC output says:
```
checking for /usr/libexec/clang-analyzer/scan-build/ccc-analyzer option to accept ISO C99... none needed
```
when configuring using the ccc-analyzer. I set -std=c99 explicitly in my projects, so I've never seen this issue before.

spbnick:
Except for the comments, this is fine. Thank you, Daniel! Only perhaps keeping commit subjects shorter would be better, perhaps 50-60 characters?

Ah, and please put the fixes *before* the integration patch, so CI keeps passing and we don't (theoretically) mask other possible issues.

spbnick:
Could we just add CFLAGS="-std=C99" to the analyzer run environment to workaround that configure/clang issue? We don't need to do it now, but perhaps later, if C99 becomes unavoidable.

dkopecek:
@spbnick I'm all in for C99 flag in CFLAGS, not just obscured in the .travis.yml to hack around the AC_PROG_CC_STDC or _C99 autodetection stuff, but simply as a global flag in the project CFLAGS.

dkopecek:
@spbnick btw. I've removed the NULL pointer guards from the commit.

spbnick:
Yeah, perhaps we can simply go for CFLAGS. It doesn't seem like we're going to be building with anything besides gcc/clang.

dkopecek:
@spbnick I'm not sure whether I understand the ordering vs. masking argument. Ideally, the fixes would be submitted as separate PRs, because this PR should introduce only the extension to CI testing which uncovered some bugs. Fixing them in the history earlier doesn't seem logical to me.

dkopecek:
@spbnick What about the optarg handling? Do you want me to fix that or are you going to refactor the code after merge?

spbnick:
@dkopecek Well, the idea is to keep all commits passing CI so the history is easier to reason about and things like bisecting work. In the current order there would be three commits failing CI, whereas if we reorder them there will be none.

spbnick:
Regarding the optarg, IMO it's better to keep number of commits to a minimum required (but not less) and reduce noise in the history. If you don't have the time or desire to mess with them now, I can redo it myself and create a new PR or modify them locally and then push directly.

spbnick:
Also by keeping CI working where we can, we let it run *all* checks always. I.e. if we can avoid CI breakage, we need to do that so all checks run and it doesn't turn out that one of the checks would fail if it did run (i.e. the issue was "masked" by an avoidable failure), and later would pop out when, say, the commit is picked/merged into another branch.

dkopecek:
@spbnick ok, the commits are reordered. I'm done with tlog for today, so either I do it on Monday or feel free to make the fixes yourself. Thanks, I learned something today!:-)

spbnick:
Thanks, Daniel :) I'll try to find some time for this on the weekend (have to compensate for some lost time anyway).

spbnick:
Moved to #52.

PR #52, opened 2016-03-12
Run CI with more build configurations, plus clang-analyzer

spbnick:


spbnick:
@dkopecek, right and there I thought Travis CI can run tests on *every* commit in the pull requests. So far I see that this is not happening and couldn't find a confirmation Travis CI can actually do that. Do you know?

This makes the need to have every commit passing and have good ordering kind of less important, but I would still like to try to have it.

Now regarding my changes from your pull request, I redid the absent optarg handling, and made the commit subjects shorter. I also changed the `[travis-ci]` tag into just `ci`. First we don't have any other CI yet to warrant mentioning Travis, then we can save subject space by using just the colon. This also matches other projects, including the Linux kernel, better.

dkopecek:
@spbnick travis-ci is not per-commit. It could be if you pushed or created a PR per commit, but that is overkill.

spbnick:
@dkopecek, oh well, this is better than nothing anyway :)

PR #53, opened 2016-03-14
CI: Enable coverage testing

dkopecek:
- Enable coverage testing by installing the lcov utility inside the
   travis sandbox.
 - Limit the coverage testing to GCC builds only as lcov doesn't seem
   to be able to process gcda files from a clang compiled project.
 - Add a coverage badge to the main README.md file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
 - CONFIGURE_ARGS="" DISTCHECK=yes
 - CONFIGURE_ARGS="--enable-debug" DISTCHECK=yes
 - CONFIGURE_ARGS="--disable-debug" DISTCHECK=yes
-- COVERAGE=yes CFLAGS="--coverage" LDFLAGS="--coverage"
+- COVERAGE=yes CFLAGS="--coverage" LDFLAGS="--coverage" CONFIGURE_ARGS="--enable-debug"
 - CLANG_SCAN_BUILD="scan-build --status-bugs"
 compiler:
 - clang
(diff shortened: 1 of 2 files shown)


dkopecek:
Do not merge this yet. There are still some problems with the coverage testing:
```
Processing src/tlog-test-fd-json-reader.gcda
Out of memory!
```

dkopecek:
Is seems this is happening only for the clang build. gcda files from GCC compiled build are processed without an error. There's no point in gathering the coverage information from two different compilers anyway, so I'll make sure that we do it only when building with GCC.

dkopecek:
Ok, looks like it's working.

spbnick:
Thank you, Daniel, this looks good!
Yes, lcov is picky about the toolchain. Shouldn't llvm-cov be used instead with clang?
In any case, yes, we don't need coverage data from both compilers.
However, can we run coverage build with "--enable-debug"?

dkopecek:
@spbnick Does the test suite test more things with --enable-debug?

spbnick:
@dkopecek Well, assertions get executed and so are the validation functions.

dkopecek:
@spbnick I understand that this is good for validation, but does it make sense in the context of coverage testing?

dkopecek:
I'll try to enable it and lets see what happens.

spbnick:
@dkopecek I'm not sure, but at least we'll get to know which validation functions aren't getting called, and either call them or remove them. Plus it will increase the coverage score, so we don't have to check coverage of files, which have low coverage only because of validation functions.

dkopecek:
Looks good

spbnick:
Splendid :)!

PR #108, opened 2017-07-10
Added a rate-limiting feature to log writing

### LINKED ISSUE #96: Limiting input collection based on rate
There's a huge amount of data collected with tlog, which is really thorough, but it fills logs with a lot of "unnecessary" data (i.e. someone holding pagedown on a man page or holding backspace). 
I'm trying to figure out how to limit the collection to "necessary" input by predicting someone is doing something based on rate of typing, similar to in [sudosh2](https://github.com/onyxpoint/sudosh2/blob/master/src/sudosh.c), but I'm not quite sure where to look in tlog to begin. As a kind of starting point for this: where does tlog actually look for and/or collect the data for a process like tlog-rec-session?

bec-f:
Added a configurable feature to limit what entries go into the log based on the rate of output. It can be disabled, and when enabled, the “cutoff rate” for collection can be changed if wanted/needed using the configuration file.

I was able to make a validate statement for the configuration rate object in lib/rec_session_conf_validate.c.

Specific changes made:
* Added code (in comments) to src/tlog-rec-session.c, lib/rec.c, lib/source.c, and lib/tty_source.c to be able to trace through the code and see the rate of each “group” collected and whether it was written to the sink or not.
    -Helps for debugging or determining rates
    -NOTE: The current location for making a trace file is in /var/lib/tlog, which could potentially create very large files in that space. Make sure that file is created and has mode 756 to be able to use it
* In lib/rec.c:
    -In tlog_rec_transfer:
        *Added two new parameters: bool use_rate (true if rate-limiting should be enabled); double cutoff_rate (If rate-limiting is enabled, only allow rates smaller than this one to be written)
        *Added new variables: int Bytes_read (accumulating total of bytes read by tlog_rec_transfer; doesn't include sizes of non-written packets); int Bytes_written (accumulating total of bytes written to the sink/log before each flush of the sink); double cur_rate (the reading rate of the current packet)
        *Added timing calculations around the reading call to use for calculating the rate, as well as accumulating the total number of bytes read after each read
        *Before writing an I/O packet to the sink, calculated the rate using the packet size and the rate at which it was collected. If rate-limiting is enabled and that packet's rate is higher than the “set rate,” then that write is skipped
        *When flushing the sink, resets Bytes_written back to 0
    -In tlog_rec:
        *Added new variables for determining use of rate
        *Added variables to the tlog_rec_transfer call
* In lib/conf_validate.c.m4:
    -Added validation code for the rate value (validate the double value, with an upper and lower bound), using the latency/payload code blocks as a reference
* In m4/tlog/rec_common_conf_schema.m4:
    -Added an M4_TYPE_DOUBLE rate variable for the configuration file
* In m4/tlog/conf_cmd.m4:
    -Added `pushdef`s for type M4_TYPE_DOUBLE
* In m4/tlog/conf_schema.m4:
    -Added a type for M4_TYPE_DOUBLE
* Added a file 'rate_config.txt' to contain the information about the rate-limiting feature

Closes #96


spbnick:
@bec-f, thank you for looking into this and contributing your time.

Regarding the floating point numbers. Do you really need them? Can you just use a smaller units and switch to integers? That should make things simpler and the change smaller. Sorry, I'm not sure yet what your units are at the moment.

How willing would you be to work on bringing this closer in line with the project's design and conventions? We'll need to move this into a separate module and plug it into the code a little differently and have some adjustments to the code style. If you wouldn't want to do this and would like to contribute as is, I can do this, but I might not find the time for this for a while. OTOH, if you'd like to help with that, I'll be able to guide you.

Thanks again for work!

spbnick:
Anyway, I'll try to understand what exactly you're doing and we'll make it work first :)

bec-f:
I did figure out how to fix the problem with printing floating point numbers (just needed to look at your other m4 code a little closer). I can look into making the units smaller. Right now they're the way they are due to the fact that after some digging, what I currently have is the only solution I found that worked in small enough time changes to be able to calculate a rate. I believe it's bytes/millisecond at the moment, but I'll double-check on that and look into using integers instead.

I don't think I have a good understanding of what you mean in the second paragraph. Could you elaborate a little bit more when you get the chance?  I'd be happy to keep working on it for the time being!

bec-f:
Sorry! I didn't mean to close it! I hit the wrong button by accident.

spbnick:
Sorry, had a very busy day today. Will respond first thing tomorrow morning.

spbnick:
I went through your changes and have a bunch of comments and requests for the next round. There will be some more later, but this will do for the start.

First, let's move all the rate limiting code to a separate module, a new type of sink, similar to `json_sink` and `tty_sink`. That will let us separate it easily and keep the logic from creeping into other parts unnecessarily. That sink should accept another sink as part of its creation parameters, and should write the data that passes the rate limit cut-off to that other sink.

This way we will simply be able to create the rate limit sink with configured parameters in `tlog_rec_create_log_sink` and put it on top of the JSON sink, if rate limiting is enabled. I think there is no need to modify the core transfer logic in `tlog_rec_transfer`. So far I also don't see the need to do extra `clock_gettime` calls, as we already have data timestamps in packets.

If we switch cutoff rate to integers, it might be easier to get rid of the extra boolean parameter that enables rate limit. We might be able to use value zero as "no rate limit", because it is useless for anything else.

Please don't make variables global unnecessarily. `Bytes_read` and `Bytes_written` can be local to `tlog_rec_transfer`.

I also have some requests for you which will help me review your changes faster as you work on them, and save time for both of us.

Please leave out any debugging/tracing code before submitting the patches, as it makes it harder to notice and read the actual logic. If you'd like to still have that tracing code locally, make it a separate commit below your main changes, and rebase the actual changes to master, dropping that commit in a branch, before pushing. I can help you with figuring out that last part.

Please also don't leave comments about what was changed in the code, as that becomes obvious when checking diffs. Please describe your changes in the commit message, instead, but don't go into details which can be easily seen in the diff. If you have to make it clear, just explain what code does in a comment, not what was added.

Please do not introduce irrelevant changes, such as changes to indentation, formatting, or amount of whitespace and blank lines, they also make it harder to notice actual changes and distract from reading the code.

I had to go through your code and revert most of the above so that I could see and concentrate on the important stuff.

Please also make sure to rebase your changes on top of latest master for each new push so it's easier for me to see your changes.

Thank you, and don't hesitate to write, if you have any questions.

bec-f:
First off, thanks for all the comments about formatting/making commits and pull requests. I'm still fairly new to GitHub, so I completely forgot that the diffs show changes.  I use the add/update comments for myself while I'm developing as well, so I'll be sure to take those comments out before I commit.

When you say to make a new sink, do you mean making a new `X_sink.c` file similar to those other two or just adding on to another sink? I haven't worked extensively with sinks before, but I can give it a shot.

As for the other changes to `tlog-rec_transfer`, I added the clocks to be able to get the exact time for how long it took to read in the data (or make the packet) in order to calculate the rate of how fast the data was read. Do the timestamps hold how long it took to read in the data? My understanding was that they only held when the packet was made, not how long it took to make the packet.  If I leave the clocks in, I could also pass the time elapsed into the new sink (as a parameter) and use that and the packet size to calculate and determine whether the packet goes through or not.

I'm working on the integers (they're huge--in the hundreds of thousands for bytes per second), and that's a good call about the boolean. I'll do that too. The global variables are taken care of now.

spbnick:
Yeah, usually open-source developers look directly at diffs when reviewing changes, and that's why a diff is a product in itself, and needs to look clean and nice, when contributing.

I would recommend trying to use `git diff` to look at your own changes, and instead of writing what you change in the code, do separate commits and describe changes in the message. Then inspecting the git logs would help you understand what you did. Try `git log -p`. That's the recommended approach.

Yes, let's e.g. name it `rate_limit_sink`, and put it into `include/tlog/rate_limit_sink.h` and `lib/rate_limit_sink.c`. Follow the example of other sinks.

The timestamps are when the data was read from the pseudo-terminal. Since tlog doesn't read a new piece (packet) from the pseudo-terminal before the previous one was encoded, the time between packet timestamps includes time to encode, and occasionally time to write a full log message.

I.e. for packet A directly followed by packet B, timestamp(B) - timestamp(A) = MAX(time_to_encode(A) + time_to_write_a_message_if_full, user_write_time(B) - user_write_time(A)).

I think that should be sufficient for rate limiting. We can start with deducing elapsed time from packet timestamps for simplicity, and improve it if it's insufficient later.

Thanks!

Don't hesitate to update the pull request (force-push to the same branch) if you'd like to ask questions on your current state of code.

bec-f:
I switched everything to integers and removed the extra comments.

I did start trying to make the rate-limiting a new sink module, but I encountered some problems in doing so. I need to able to organize my thoughts and am about to leave, so I will let you know tomorrow!   I was planning to push the start that I had made into a different branch, but in modifying my code to push here, I seem to have lost it.  I will see if I can find (or recreate) it again tomorrow.

spbnick:
Thanks! If you ever committed your lost changes, you might still find them in `git reflog`.

bec-f:
Unfortunately, I apparently hadn't committed any of the changes I lost, so I recreated them today and pushed my attempt another branch on my profile. You can look at what I have so far on this front! 
As far as the issues I've encountered, I'll list them here.

I wasn't sure how to calculate the rate with the information about the packets that you suggested. Specifically, I'm not sure how to get the information from one packet while looking at another, and from there, I'm not sure how to calculate processing/writing time (versus reading time) between each packet timestamp.

I was also not sure how to write to the log_sink from there. There are a few things I have thought of/come across in trying to make a rate-limit sink:
I integrated the log_sink with the rate-limit sink by passing it in as a parameter when the rate-limit sink is created, but writing to the log_sink in the rate-limit sink seems like it would be a more complicated version of the code that you already have in the `tlog_rec_transfer` function. 
I'm also not sure how I would modify the calls in `tlog_rec_transfer` to write to the rate-limit sink as well as keep your current error checks if I did modify it. 
Since all of the data would have to go through this sink before getting to the log_sink anyway, it might be more straightforward and possibly more efficient (less function calls and memory used) to keep the rate calculation and check where it is for now?

I see and understand where you are coming from with wanting separation of responsibility in the code, but these are some things that I've encountered in trying to do so. What do you think?

spbnick:
Yes, more code is the price to pay for modularity, but the benefit is it is easier to think of the whole and pieces in the end. I don't think it is going to be much more complicated, in total, than doing it in `tlog_rec_transfer`, at least I don't see it so far.

Could you perhaps explain what your rate limiting calculation should be, in pseudo-code, or just in English? Then maybe it will be easier for me to see if there are any difficulties.

spbnick:
Also, at this moment I don't worry so much about performance, than about being able to understand the code and about correctness. First we prove the code works, and is useful, and if we see that it doesn't perform well enough, then we can measure it and optimize where necessary. However, I would abandon modularity last.

spbnick:
You don't need to modify `tlog_rec_transfer` to add the rate-limiting sink. You modify `tlog_rec_create_log_sink` to check for the configuration after the regular sink is created, and if rate-limiting is necessary, you create the rate-limiting sink passing it the sink you already created, and return the rate-limiting sink instead of the regular sink.

bec-f:
My rate limiting calculation has been the size of the packet divided by how long it took to read that data (or make the packet, since the two go hand in hand). I positioned the clocks where they currently are in `tlog_rec_transfer` since that would give me very close to exactly how long it took to read and make the packets with very few "extra" things caught in the timing calculation.

Pseudo code would probably be:
`get time before reading;`

`read packet;`

`get time after reading;`

`time_elapsed = after - before;`

. //other things happening 

.

.

`rate = packet_size / time_elapsed;`

`writing to log;`

bec-f:
Okay! I will leave more specific questions on your comments on the code.  As far as the bigger picture, I think I understand what you mean with modifying `tlog_rec_create_log_sink`, so I will take a look there. My biggest question with writing to the log in the rate-limiting sink is how do I check for errors there (similar to how you check in `tlog_rec_transfer`). Would I do the error check there or simply return the tlog_grc and let the error checks in `tlog_rec_transfer` do the rest?

bec-f:
Also, I've never worked with Automake files. How would I add these new sink files to those in order to test the changes?

spbnick:
I see. I don't think your rate calculation above is correct (or at least it is incomplete), because you need to account for all the time passing between the reads, not just the time waiting for the new piece of data to arrive, plus you don't accumulate data on how many packets and of which size were output in the past. I.e. if you get many small packets in quick succession, which were read fast, you can easily exceed the rate. Forgive me if I misunderstood you, though. And, unfortunately, I don't have time right now to research what exactly should be done instead, it's my last working day before holiday. Still, I'm pretty sure that you have access to enough timing information within the sink.

Basically, right now the only return code that doesn't necessarily terminate the program is EINTR from libc, and you're supposed to pass it along. I don't think you need to handle any return codes specifically. Just abort your processing when you receive an error from the lower sink, and return it.

You don't need to know much about Automake files for your purposes so far. Just add your sink .h and .c files similarly to other sinks, e.g. like `tty_sink`. Keep the file lists sorted, though, please.

trevor-vaughan:
@spbnick and @bec-f I've been following this thread with a lot of interest since, as mentioned, this is the one thing keeping me from using `tlog` and ditching `sudosh2`.

The current implementation seems to work as I would expect from trying it out on some of my systems. I haven't lost any expected data and I've ditched 90% of some boneheaded things like `rpm2cpio my.rpm`.

If you get a lot of small packets in quick succession, it should exceed the rate, shouldn't it? That means that someone is probably dumping random garbage to the screen, flipping through a man page, or whatever.

bec-f:
@spbnick and @trevor-vaughan  I believe I have the new sink implemented and working!

For some reason, it still isn't adding the default rate (0) to the configuration file even though it has a rate section and the comment explaining how the rate part works. That's the only hiccup I've encountered so far, and I'm not sure what else to do because my m4 code looks like the others.

bec-f:
The checks that failed are due to warnings I got about `const` pointers. I'm going to go through and look at/fix those tomorrow.

trevor-vaughan:
Nice! I'm looking forward to checking this out.

bec-f:
I'm still working on this, since it's not passing because of errors (that didn't show up on my machine). I'll try to take a closer look tomorrow.

lib/rate_limit_sink.c
```diff
@@ -0,0 +1,152 @@
...
+        else {
+            tlog_pkt_pos_move_past(rate_limit_sink->prev_ppos, rate_limit_sink->prev_pkt);
+            return TLOG_RC_OK;
+        }
+    }
+
+end:
+    rate_limit_sink->prev_pkt = pkt;
+    rate_limit_sink->prev_ppos = ppos;
+    rate_limit_sink->prev_end = end;
```
spbnick: You're copying pointers here instead of the contents of `pkt`, `ppos`, and `end`. When you dereference those pointers next time you'll get different values, because the location would have likely changed, and possibly would read completely incorrect memory, if the original values were allocated with malloc and then freed between the calls.

Instead declare those fields as plain structures, not pointers, and copy into them here.

lib/rec.c
```diff
@@ -657,6 +686,14 @@ tlog_rec(struct tlog_errs **perrs, uid_t euid, gid_t egid,
...
     latency = (unsigned int)num;
 
+    /* Read the rate limit*/
+    if (!json_object_object_get_ex(conf, "rate", &obj)) {
+        tlog_errs_pushs(perrs, "Rate is not specified");
+        grc = TLOG_RC_FAILURE;
+        goto cleanup;
+    }
+    check_rate = json_object_get_int64(obj);
+
```
spbnick: Since you're passing `conf` to `tlog_rec_create_log_sink` anyway, there is no need to read the rate here and pass it as an extra argument, just read it in `tlog_rec_create_log_sink`.

lib/rate_limit_sink.c
```diff
@@ -0,0 +1,152 @@
...
+{
+    struct tlog_rate_limit_sink *rate_limit_sink =
+        (struct tlog_rate_limit_sink *)sink;
+    tlog_grc grc;
+
+    if(rate_limit_sink->prev_pkt == NULL) {
+        grc = TLOG_RC_OK;
+        goto end;
+    }
+    if(rate_limit_sink->prev_pkt->type != TLOG_PKT_TYPE_IO || rate_limit_sink->rate == 0){
```
spbnick: Couldn't it be that if we do a lot of window resizes we can bypass rate limit here?

lib/rate_limit_sink.c
```diff
@@ -0,0 +1,152 @@
...
+        grc = tlog_sink_write(rate_limit_sink->dsink, rate_limit_sink->prev_pkt,
+                rate_limit_sink->prev_ppos, rate_limit_sink->prev_end);
+        goto end;
+    }
+    else {
+
+        uint64_t time_change = (pkt->timestamp.tv_sec - rate_limit_sink->prev_ts.tv_sec) * 1000000 + (pkt->timestamp.tv_nsec - rate_limit_sink->prev_ts.tv_nsec)/1000;
+        /* Get the rate as an integer with decimal precision */
+        double d_rate = (double)(rate_limit_sink->prev_pkt->data.io.len)/(double)(time_change);
+        uint64_t cur_rate = d_rate * 1000000;
```
spbnick: I'm still unconvinced that this will work. Have you considered token bucket / leaky bucket algorithms?

lib/rate_limit_sink.c
bec-f: I had considered that with whether the packet data would change, but when I tested it with the pointers, it seemed to do what I needed it to. I'll admit that I had not considered malloc/free situations, so that is a good catch.

lib/rate_limit_sink.c
bec-f: My understanding could be wrong, but I didn't think it would bypass rate-limiting.  Technically, in a sense, I suppose it would since I'm not checking rate on window resizes. However, the focus was limiting data from I/O, not window resizing, so window resize rates wouldn't matter (if I'm understanding your comment/question correctly).  I also can't think of a way someone would be able to bypass rate-limiting for I/O data through window resizing since the packets seem to be sequential and not parallel.

Is that along the lines you were thinking?

lib/rec.c
bec-f: Okay! I can do that, it should be a quick/easy change.

bec-f: I've got this done, but working on all the other things.

lib/rate_limit_sink.c
bec-f: I'm actually not familiar with these algorithms, I can look into them, but the next thing I'm looking at is the pointers.

bec-f:
@sbpnick and @trevor-vaughan:  I've been able to move reading the rate and I've started on the pointers change. Unfortunately, I've been working on this as a summer project, and I'm about out of time (I only have tomorrow left).  I'll be sure to leave my work on my page and/or here so you can use it if it'd be helpful to get this ticket done!

bec-f:
@spbnick I've checked that maintainers can make updates, so you can use my code as a base if it would be useful for future development.

dpal:
What is the state of this? Is it ready for merge or still requires some cleanup?

spbnick:
Thank you, @bec-f! I hope to get some time to work on this in autumn, unless you're able to proceed.

@dpal, the main issue is proofing the rate-limiting algorithm. Some cleanup and simplification will be needed, but shouldn't be hard.

spbnick:
I've implemented rate limiting in #124.

The difference is that the limiting is done at the level of formatted log messages and full log messages are dropped only. This avoids miscalculating actual final log rate, as some invalid characters could be blown up considerably in size due to escaping, plus delays and window changes add their own overhead. It also allows detecting when something was dropped, as there would be a discontinuity in message IDs.

Whoever is interested, please take a look at #124 instead. I'm going to merge it tomorrow, if nobody objects.

I'm closing this PR, but please feel free to comment again, if you want to discuss anything regarding this specific approach.

PR #125, opened 2017-10-12
README: add a note about how to make release tarball

frasertweedale:
The README talks about using release tarballs (particularly for
generating (S)RPM packages), but does not mention how to generate a
release tarballs.  Add a note on how to do that.


spbnick:
Thanks, Fraser!

PR #128, opened 2017-10-13
Do not add tlog-rec-session to /etc/shells

spbnick:
Remove adding and removing tlog-rec-session to/from /etc/shells from
tlog.spec. This prevents chsh from allowing the users to change their
shells from tlog-rec-session.


spbnick:
Merged.

PR #137, opened 2017-11-21
update configuration raw mode in termios

danghai:
According TERMIOS(3),
cfmakeraw() sets the terminal to something like the "raw" mode of the old Version 7 terminal driver: input is available character by character, echoing is disabled, and all special processing of terminal input and output characters is disabled. The terminal attributes are set as follows:
```c
    termios_p->c_iflag &= ~(IGNBRK | BRKINT | PARMRK | ISTRIP
                    | INLCR | IGNCR | ICRNL | IXON);
    termios_p->c_oflag &= ~OPOST;
    termios_p->c_lflag &= ~(ECHO | ECHONL | ICANON | ISIG | IEXTEN);
    termios_p->c_cflag &= ~(CSIZE | PARENB);
    termios_p->c_cflag |= CS8;
```
It makes sure that you do not have any bugs in raw mode in termios. It is important because tlog_play will use raw mode in termios to record, in which terminal is the noncanonical mode, with all input and output processing, as well as echoing, switched off.

### DIFF
--- a/src/tlog-play.c
+++ b/src/tlog-play.c
@@ -513,10 +513,12 @@ run(struct tlog_errs **perrs,
      * but keep signal generation, if not persistent
      */
     raw_termios = orig_termios;
-    raw_termios.c_lflag &= ~(ICANON | IEXTEN | ECHO | (persist ? ISIG : 0));
+    raw_termios.c_lflag &= ~(ICANON | IEXTEN | ECHO | ECHONL | (persist ? ISIG : 0));
     raw_termios.c_iflag &= ~(BRKINT | ICRNL | IGNBRK | IGNCR | INLCR |
                              INPCK | ISTRIP | IXON | PARMRK);
     raw_termios.c_oflag &= ~OPOST;
+    raw_termios.c_cflag &= ~(CSIZE | PARENB);
+    raw_termios.c_cflag |= CS8;
     raw_termios.c_cc[VMIN] = 1;
     raw_termios.c_cc[VTIME] = 0;
     rc = tcsetattr(STDOUT_FILENO, TCSAFLUSH, &raw_termios);


spbnick:
Thank you, Hai. However, I don't see how ECHONL is necessary, since we have ECHO on. Also, I don't think we should change serial port settings (i.e. parity with PARENB, and character size with CSIZE). If you think otherwise, could you please explain your reasoning behind those?

danghai:
`    termios_p->c_lflag &= ~(ECHO | ECHONL | ICANON | ISIG | IEXTEN);`
I see they mention "

> If this bit is set and the ICANON bit is also set, then the newline ('\n') character is echoed even if the ECHO bit is not set."

 from https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_chapter/libc_17.html#SEC361 . However, the ICANOT turn off, ECHONL does not work too. So, I think ECHONL does not matter here . You can ignore the ECHONL. 

```
    termios_p->c_cflag &= ~(CSIZE | PARENB);
    termios_p->c_cflag |= CS8;
```
The CSIZE masking is used to reset the data bits field to zeros. This allows the CS8 bit pattern to be or -ed in later, setting the data bits value to 8 bits. Disabling the PARENB flag causes parity generation on output to be disabled, and disables parity checking on input. 
From their link , they metion about PARENB.

>  If this bit is not set, no parity bit is added to output characters, and input characters are not checked for correct parity.

spbnick:
Yeah, I understand what those bits do, but I don't see why we should be changing them. Serial port should be setup by system administrator, in /etc/inittab or systemd units. We have no business changing them. The other side of the serial line might be set up differently, e.g. with parity enabled, however unlikely nowadays.

danghai:
Yeah you are right, I read some parts in termios. It does not have effect on the terminal after recording. as I said before, when I read the Termios part carefully, they mention about raw mode setting, I see you do not set it. Actually, it does not have effect on your tlog.

spbnick:
Alright, I'll close this PR for now, but if you find a case where tlog breaks without this, please reopen. Thank you!

Issue #30, opened 2016-02-26
Check timing delay range is within TLOG_DELAY_MAX in source


spbnick:

spbnick:
This basically means adding assertions to various tlog sources, including the abstract interface, that the successively returned packets are not more than TLOG_DELAY_MAX apart. Any associated data or initialization code should be compiled in debug build only, i.e. be under `#ifndef NDEBUG` conditions.

danghai:
When I look the man page for `assert` . 
       If  the  macro  NDEBUG  is  defined  at  the moment <assert.h> was last
       included, the macro assert() generates no code, and hence does  nothing
       at  all.   It  is not recommended to define NDEBUG if using assert() to
       detect error conditions since the software may behave non-deterministi‐
       cally.

And I am not sure, why do you need the NDEBUG. Can you explain more? Do you use it for DEBUG option? I mean if it violates the TLOG_DELAY_MAX, it will give the message about it and abort? 
I have another question about `delay.h`. I see it is inside lib but you never include it. Do you use the macro NDEBUG in it?

spbnick:
I'm using `assert` with NDEBUG as a compromise. With assert I only check conditions which can arise from internal, coding errors (at least that's the intention). All other, outside errors, e.g. from OS conditions, user input, etc. are checked with regular conditions and error handling.

I do testing without NDEBUG defined (with --enable-debug option specified), i.e. with `assert` invocations intact, but I define NDEBUG in production builds (without --enable-debug) to improve performance, as all the checks through the code are slowing it down. If a distro packager is OK with reduced performance, they might choose to not define NDEBUG for their build. It's a debatable practice, but one that I follow for now.

If you do not have NDEBUG defined, e.g. if you configured the build with the --enable-debug option, and your `assert` invocations are in effect, then when one fails, a message will be printed and the process will be killed by SIGABRT, producing a coredump, if that's enabled.

If you'd like, write a small program using `assert` and try building and running it with and without NDEBUG defined and assertion failing.

delay.h is in include/tlog, and it's included into seven files. You can find them as such: `grep -r --include '*.[hc]' delay.h *`.

danghai:
I understand what you mean about NDEBUG or DEBUG through building and running a simple example.
```
#include <stdio.h>
//# define NDEBUG
#include <assert.h>

int main(){
    assert(1==2);
    return 0;
}
```

[...]

PR #148, opened 2017-12-08
Check timing delay range is within TLOG_DELAY_MAX in source

danghai:
Pull request for this issue: https://github.com/Scribery/tlog/issues/30

### DIFF
--- a/include/tlog/source.h
+++ b/include/tlog/source.h
@@ -32,7 +32,10 @@
 
 /** Abstract source */
 struct tlog_source {
-    const struct tlog_source_type *type;   /**< Type */
+    const struct tlog_source_type *type;           /**< Type */
+#ifndef NDEBUG
+    struct timespec                last_timestamp; /**< Last timestamp */
+#endif
 };
 
 /**
--- a/lib/source.c
+++ b/lib/source.c
@@ -28,6 +28,10 @@
 #include <tlog/rc.h>
 #include <tlog/misc.h>
 #include <tlog/source.h>
+#ifndef NDEBUG
+#include <tlog/timespec.h>
+#include <tlog/delay.h>
+#endif
 
 tlog_grc
 tlog_source_create(struct tlog_source **psource,
@@ -101,6 +105,9 @@ tlog_grc
 tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 {
     tlog_grc grc;
+#ifndef NDEBUG
+    struct timespec diff;
+#endif
     assert(tlog_source_is_valid(source));
     assert(tlog_pkt_is_valid(pkt));
     assert(tlog_pkt_is_void(pkt));
@@ -109,6 +116,14 @@ tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 
     assert(grc == TLOG_RC_OK || tlog_pkt_is_void(pkt));
     assert(tlog_source_is_valid(source));
+#ifndef NDEBUG
+    if (!tlog_pkt_is_void(pkt)) {
+        tlog_timespec_sub(&pkt->timestamp, &source->last_timestamp, &diff);
+        assert(tlog_timespec_cmp(&diff, &tlog_timespec_zero) >= 0);
+        assert(tlog_timespec_cmp(&diff, &tlog_delay_max_timespec) <= 0);
+        source->last_timestamp = pkt->timestamp;
+    }
+#endif
     return grc;
 }
 


include/tlog/source.h
```diff
@@ -32,7 +32,8 @@
 
 /** Abstract source */
 struct tlog_source {
-    const struct tlog_source_type *type;   /**< Type */
+    const struct tlog_source_type *type;          /**< Type */
+    struct timespec               last_timestamp; /**< Timestamp */
```
spbnick: Could you please align the parameter names? I.e. add a space before `last_timestamp`?

lib/source.c
```diff
@@ -28,6 +28,8 @@
 #include <tlog/rc.h>
 #include <tlog/misc.h>
 #include <tlog/source.h>
+#include <tlog/timespec.h>
+#include <tlog/delay.h>
```
spbnick: Could you please wrap these includes in `#ifndef NDEBUG`? We don't use them otherwise here.

lib/source.c
```diff
@@ -101,6 +103,7 @@ tlog_grc
 tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 {
     tlog_grc grc;
+    struct timespec diff;
```
spbnick: Could you please wrap this line in `#ifndef NDEBUG`? Otherwise the compiler will complain this variable is unused when built with `NDEBUG`.

danghai: Oh I already put this line in `#ifndef NDEBUG` in commit issue, and you told me put this line on the top. It makes me little confused about it because I know the compiler will complain this when variable is unused when built with `NDEBUG`. I will wrap this lane and 2 #include in `NDEBUG`.

include/tlog/source.h
danghai: Do you mean a space between timespec and last_timestamp? I think *type and last_timestamp are same type ( variable) that is reason why I align these parameter names

danghai:
Thanks, I wrap `diff` and 2 #include in NDEBUG, and align `last_timestamp`.

include/tlog/source.h
```diff
@@ -33,6 +33,7 @@
 /** Abstract source */
 struct tlog_source {
     const struct tlog_source_type *type;   /**< Type */
+    struct timespec last_timestamp;        /**< Timestamp */
```
spbnick: Ah, no I meant like this:

```C
/** Abstract source */
struct tlog_source {
    const struct tlog_source_type  *type;           /**< Type */
    struct timespec                 last_timestamp; /**< Timestamp */
```

Notice how `type` and `last_timestamp` names are aligned on the same column.
Take a look at other similar structure declarations.

lib/source.c
```diff
@@ -29,6 +29,7 @@
 #include <tlog/misc.h>
 #include <tlog/source.h>
 
+
```
spbnick: Nah, not an empty line here, I was talking about that space in front of `last_timestamp`.

include/tlog/source.h
spbnick: Oh, and wrap this one in `#ifndef NDEBUG`, and fixup the comment too:

```C
/** Abstract source */
struct tlog_source {
    const struct tlog_source_type  *type;           /**< Type */
#ifndef NDEBUG
    struct timespec                 last_timestamp; /**< Last timestamp */
#endif
```

danghai:
Thanks, I forget about putting `ifndef` in file header because `last_timestamp` uses when `NDEBUG` is not defined.

lib/source.c
```diff
@@ -109,6 +109,15 @@ tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 
     assert(grc == TLOG_RC_OK || tlog_pkt_is_void(pkt));
     assert(tlog_source_is_valid(source));
+#ifndef NDEBUG
+#include <tlog/timespec.h>
+#include <tlog/delay.h>
+    struct timespec diff;
```
spbnick: Please get those into their places too. Thanks!

danghai: Do you mean put these line in Header file and including `diff` variable?

spbnick: Just imagine where you would put them *without* `#ifndef NDEBUG` (forget about it for a moment), and where you in fact already put them before, then simply wrap them in that `#ifndef` right in place.

danghai:
I think I simply put them on the top when I declare the library with `ifndef NDEBUG`.

include/tlog/source.h
```diff
@@ -32,7 +32,10 @@
 
 /** Abstract source */
 struct tlog_source {
-    const struct tlog_source_type *type;   /**< Type */
+    const struct tlog_source_type *type;           /**< Type */
+#ifndef NDEBUG
+    struct timespec                last_timestamp; /**< Timestamp */
```
spbnick: Since we have comments here, it would be better to make them at least match the specificity of the variable name. I.e. could you please write "Last timestamp" here instead of just "Timestamp"?

lib/source.c
```diff
@@ -109,6 +113,13 @@ tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 
     assert(grc == TLOG_RC_OK || tlog_pkt_is_void(pkt));
     assert(tlog_source_is_valid(source));
+#ifndef NDEBUG
+    struct timespec diff;
```
spbnick: Could you please move the declaration of `diff` to the top of the function, joining `grc` there?

danghai: Hi Nick, you want to take 2 assert in `NDEBUG` define. If I remove the `struct timespec diff`, compiler will ignore all asserts below?

spbnick: I'm sorry, I don't quite understand your question. I don't want to "take 2 assert in NDEBUG define", I'm just asking you to move the `struct timespec diff` declaration to the top of the `tlog_source_read` function, wrapping it in `#ifndef NDEBUG`, and putting it beside the `tlog_grc grc` declaration. That would make it conform to the tlog conventions better, as tlog usually has all declarations at the start of a function.

danghai:
I write "Last timestamp" in comment and put the `diff` on the top of the function

lib/source.c
danghai: Sorry I make you confused, I just want to make sure if  I write the code like: 
```
#ifndef NDEBUG
assert()
assert()
```
It will ignore all assert below. 

If I write something like: 
```
#ifndef NDEBUG
struct timespec diff
assert()
assert()
```
It will run when NDEBUG is not define

lib/source.c
```diff
@@ -109,6 +116,12 @@ tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 
     assert(grc == TLOG_RC_OK || tlog_pkt_is_void(pkt));
     assert(tlog_source_is_valid(source));
+#ifndef NDEBUG
+    tlog_timespec_sub(&pkt->timestamp, &source->last_timestamp, &diff);
+    assert(tlog_timespec_cmp(&diff, &tlog_timespec_zero) >= 0);
+    assert(tlog_timespec_cmp(&diff, &tlog_delay_max_timespec) <= 0);
+    source->last_timestamp = pkt->timestamp;
```
spbnick: One last thing: this needs to be checked only when `pkt` is not "void". Otherwise `pkt->timestamp` can be invalid.

danghai: I see the `tlog_pkt_is_void(pkt)` already check before `ifndef`. It means `pkt->timestamp` will be valid

danghai:
Hi Nick, I got it. I add check `tlog_pkt_is_void(pkt)` .  However, I put the `assert()` below `NDEBUG`, does the macro assert() generates no code?

spbnick:
Hi Hai,
If you put `assert` within `#ifndef NDEBUG`, then it will work the same as if you didn't put it there, because its definition is dependent on the same condition. Any other code within the same `#ifndef NDEBUG` would not be compiled if `NDEBUG` is defined. I know this can be confusing, but you need to figure out how preprocessor works, so I don't have to explain it every time :)

Regarding the code, you put an assert to check a condition. However, the condition you put there would make the code terminate with debugging enabled, because the assertion would fail, with a perfectly valid situation. Having a "void" packet come there is perfectly valid. It's just that we need to apply the checking that you just added only to "non-void" packets.

danghai:
Hi Nick, yeah I feel `NDEBUG` is confused both name and the way to use. I am figuring how preprocessor works and write some simple C code. 

Sorry, I do not notice about the expression in assert. I understand it makes sure that `pkt` must be not void to this expression `pkt->timestamp` is valid.

lib/source.c
```diff
@@ -109,6 +116,13 @@ tlog_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
 
     assert(grc == TLOG_RC_OK || tlog_pkt_is_void(pkt));
     assert(tlog_source_is_valid(source));
+#ifndef NDEBUG
+    assert(!tlog_pkt_is_void(pkt));
```
spbnick: This is still using `assert`, which won't work. Think how you can make the below code check the condition when the packet is not "void", and not do any checking when it is, but without terminating the program.

danghai:
I think I still use `assert`, but I put above the `NDEBUG`. 

```
     assert(!tlog_pkt_is_void(pkt));
#ifndef NDEBUG
     tlog_timespec_sub(&pkt->timestamp, &source->last_timestamp, &diff);
     assert(tlog_timespec_cmp(&diff, &tlog_timespec_zero) >= 0);
     assert(tlog_timespec_cmp(&diff, &tlog_delay_max_timespec) <= 0);
     source->last_timestamp = pkt->timestamp;
```

spbnick:
Hi Hai,

No, you shouldn't use `assert` for this at all. Putting it outside `NDEBUG` wouldn't make any difference whatsoever. Try to think this through. Ask me questions. This is easier than you think :)

danghai:
If I cannot use `assert`. I only have one way that uses `if condition` to check and handle this. It looks like: 
```
if (tlog_pkt_is_void(pkt)
     return grc;
#ifndef NDEBUG
     tlog_timespec_sub(&pkt->timestamp, &source->last_timestamp, &diff);
     ...
```

spbnick:
This is basically what we need, only in this case it's better to revert the condition and put the checking code into the true branch body for the `if` statement. This way the reasoning would be more apparent. I.e. it would say: "if packet is not void check this".

danghai:
I revert the condition and put the checking code into the true branch body of the `if` statment.

spbnick:
OK, good. However, now we get an `if` with an empty body if `NDEBUG` is defined. Can you fix that?

danghai:
Yeah, I put `if` condition in `NDEBUG`

spbnick:
Great, now everything is right. Thank you, Hai!

PR #153, opened 2017-12-20
Implement error handling macros

### LINKED ISSUE #116: Implement error handling macros
Throughout the code several error-handling patterns are repeated many times. Implement macros to handle the most typical cases and save on line count.

danghai:
Fixes #116 
I send the first draft. As I said before, I implement error handling macros based on 4 types that you use. I feel it is not good idea to do that. Could you give some advice? 

> For type 1: 

```
        if (!json_object_object_get_ex(conf, "file", &conf_file)) {
            tlog_errs_pushs(perrs,
                            "File reader parameters are not specified");
            grc = TLOG_RC_FAILURE;
            goto cleanup;
        }
```
--> It becomes: 
```
        if (!json_object_object_get_ex(conf, "file", &conf_file)) {
            grc = tlog_handler1(perrs, TLOG_RC_FAILURE, "File reader parameters are not specified");
            goto cleanup;
        }
```

> For type 3: 

```
    if (setlocale(LC_ALL, "") == NULL) {
        grc = TLOG_GRC_ERRNO;
        tlog_errs_pushc(&errs, grc);
        tlog_errs_pushs(&errs,
                        "Failed setting locale from environment variables");
        goto cleanup;
    }
```
---> It becomes: 

```
    if (setlocale(LC_ALL, "") == NULL) {
        grc = tlog_handler3(&errs, TLOG_GRC_ERRNO, "Failed Setting locale from environment variables")
        goto cleanup;
    }
```
I do not finish function relating `pushf` because of its complex argument, and I wait for your thought about implementation. Do you implement all these situation in a function?

### DIFF
--- a/include/tlog/errs.h
+++ b/include/tlog/errs.h
@@ -141,4 +141,64 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
  */
 extern int tlog_errs_print(FILE *stream, const struct tlog_errs *errs);
 
+/**
+ * Push an error to the "perrs" stack, with message as a constant string
+ * and then go to cleanup.
+ *
+ * @param _msg   The message string to copy and put into the pushed error.
+ *               Cannot be NULL.
+ */
+#define TLOG_ERRS_RAISES(_msg) \
+     do {                                                                 \
+         tlog_errs_pushs(perrs, _msg);                                    \
+         goto cleanup;                                                    \
+     } while (0)
+
+/**
+ * Push two errors to the "perrs" stack, first one with message being a global
+ * return code description, and the second with message as a constant string.
+ * And then go to cleanup.
+ *
+ * @param _grc  The global return code, which description will be pushed to
+ *              the stack.
+ * @param _msg  The message string to copy and put into the pushed error.
+ *              Cannot be NULL.
+ */
+#define TLOG_ERRS_RAISECS(_grc, _msg) \
+     do {                                                                 \
+         tlog_errs_pushc(perrs, _grc);                                    \
+         tlog_errs_pushs(perrs, _msg);                                    \
+         goto cleanup;                                                    \
+     } while (0)
+
+/**
+ * Push an error to the "perrs" stack, with a sprintf-formatted message.
+ * And then go to cleanup.
+ *
+ * @param _fmt  The message sprintf(3) format string. Cannot be NULL.
+ * @param ...   The message format arguments.
+ */
+#define TLOG_ERRS_RAISEF(_fmt, _args...) \
+     do {                                                                 \
+         tlog_errs_pushf(perrs, _fmt, ##_args);                           \
+         goto cleanup;                                                    \
+     } while (0)
+
+/**
+ * Push two errors to the "perrs" stack, first one with message being a global
+ * return code description, and the second with a sprintf-formatted message.
+ * And then go to cleanup.
+ *
+ * @param _grc  The global return code, which description will be pushed to
+ *              the stack.
+ * @param _fmt  The message sprintf(3) format string. Cannot be NULL.
+ * @param ...   The message format arguments.
+ */
+#define TLOG_ERRS_RAISECF(_grc, _fmt, _args...) \
+     do {                                                                 \
+         tlog_errs_pushc(perrs, _grc);                                    \
+         tlog_errs_pushf(perrs, _fmt, ##_args);                           \
+         goto cleanup;                                                    \
+     } while (0)
+
 #endif /* _TLOG_ERRS_H */
--- a/lib/tlog/session.c
+++ b/lib/tlog/session.c
@@ -93,51 +93,41 @@ cleanup:
  * In case of an error setting/restoring EUID/EGID, sets the "grc" variable,
  * pushes messages to the specified error stack and jumps to "cleanup" label.
  *
- * @param _perrs    Location for the error stack. Can be NULL.
  * @param _euid     The EUID to set temporarily.
  * @param _egid     The EGID to set temporarily.
  * @param _expr     The expression to evaluate with EUID/EGID set.
  */
-#define EVAL_WITH_EUID_EGID(_perrs, _euid, _egid, _expr) \
-    do {                                                        \
-        struct tlog_errs **__perrs = (_perrs);                  \
-        uid_t _orig_euid = geteuid();                           \
-        gid_t _orig_egid = getegid();                           \
-                                                                \
-        /* Set EUID temporarily */                              \
-        if (seteuid(_euid) < 0) {                               \
-            grc = TLOG_GRC_ERRNO;                               \
-            tlog_errs_pushc(__perrs, grc);                      \
-            tlog_errs_pushs(__perrs, "Failed setting EUID");    \
-            goto cleanup;                                       \
-        }                                                       \
-                                                                \
-        /* Set EGID temporarily */                              \
-        if (setegid(_egid) < 0) {                               \
-            grc = TLOG_GRC_ERRNO;                               \
-            tlog_errs_pushc(__perrs, grc);                      \
-            tlog_errs_pushs(__perrs, "Failed setting EGID");    \
-            goto cleanup;                                       \
-        }                                                       \
-                                                                \
-        /* Evaluate */                                          \
-        _expr;                                                  \
-                                                                \
-        /* Restore EUID */                                      \
-        if (seteuid(_orig_euid) < 0) {                          \
-            grc = TLOG_GRC_ERRNO;                               \
-            tlog_errs_pushc(__perrs, grc);                      \
-            tlog_errs_pushs(__perrs, "Failed restoring EUID");  \
-            goto cleanup;                                       \
-        }                                                       \
-                                                                \
-        /* Restore EGID */                                      \
-        if (setegid(_orig_egid) < 0) {                          \
-            grc = TLOG_GRC_ERRNO;                               \
-            tlog_errs_pushc(__perrs, grc);                      \
-            tlog_errs_pushs(__perrs, "Failed restoring EGID");  \
-            goto cleanup;                                       \
-        }                                                       \
+#define EVAL_WITH_EUID_EGID(_euid, _egid, _expr) \
+    do {                                                                 \
+        uid_t _orig_euid = geteuid();                                    \
+        gid_t _orig_egid = getegid();                                    \
+                                                                         \
+        /* Set EUID temporarily */                                       \
+        if (seteuid(_euid) < 0) {                                        \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed setting EUID");               \
+        }                                                                \
+                                                                         \
+        /* Set EGID temporarily */                                       \
+        if (setegid(_egid) < 0) {                                        \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed setting EGID");               \
+        }                                                                \
+                                                                         \
+        /* Evaluate */                                                   \
+        _expr;                                                           \
+                                                                         \
+        /* Restore EUID */                                               \
+        if (seteuid(_orig_euid) < 0) {                                   \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed restoring EUID");             \
+        }                                                                \
+                                                                         \
+        /* Restore EGID */                                               \
+        if (setegid(_orig_egid) < 0) {                                   \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed restoring EGID");             \
+        }                                                                \
     } while (0)
 
 tlog_grc
@@ -156,9 +146,7 @@ tlog_session_lock(struct tlog_errs **perrs, unsigned int id,
     /* Format lock file path */
     grc = tlog_session_format_lock_path(&path, id);
     if (grc != TLOG_RC_OK) {
-        tlog_errs_pushc(perrs, grc);
-        tlog_errs_pushs(perrs, "Failed formatting lock file path");
-        goto cleanup;
+        TLOG_ERRS_RAISECS(grc, "Failed formatting lock file path");
     }
 
     /*
@@ -166,7 +154,7 @@ tlog_session_lock(struct tlog_errs **perrs, unsigned int id,
      * Assume session ID never repeats (never wraps around).
      * FIXME Handle repeating session IDs.
      */
-    EVAL_WITH_EUID_EGID(perrs, euid, egid,
+    EVAL_WITH_EUID_EGID(euid, egid,
                         fd = open(path, O_CREAT | O_EXCL, S_IRUSR | S_IWUSR));
     if (fd < 0) {
         int open_errno = errno;
@@ -212,13 +200,11 @@ tlog_session_unlock(struct tlog_errs **perrs, unsigned int id,
     /* Format lock file path */
     grc = tlog_session_format_lock_path(&path, id);
     if (grc != TLOG_RC_OK) {
-        tlog_errs_pushc(perrs, grc);
-        tlog_errs_pushs(perrs, "Failed formatting lock file path");
-        goto cleanup;
+        TLOG_ERRS_RAISECS(grc, "Failed formatting lock file path");
     }
 
     /* Remove the lock file */
-    EVAL_WITH_EUID_EGID(perrs, euid, egid, rc = unlink(path));
+    EVAL_WITH_EUID_EGID(euid, egid, rc = unlink(path));
     if (rc < 0) {
         int unlink_errno = errno;
         tlog_grc unlink_grc = TLOG_GRC_ERRNO;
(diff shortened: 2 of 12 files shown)


spbnick:
Thank you, Hai. Hmm, I like the fact that you're using inline functions, as it's a good practice to use them instead of macros wherever possible. However, we might still need macros if we are to seriously minimize the error-handling code.

One thing from the start is we'll need meaningful names for those macros/functions, and not just "handler".
Another thing is that tlog (and my other projects) follow a convention where symbols in a module start with module name. E.g. note how all the symbols defined in `tlog_errs.h` start with `tlog_errs_`, same with other modules. Your module will need to do the same.

Now, even though it's common knowledge that your macros should not implicitly access any variables, or change the control flow of your program, I don't think we can achieve significant code reduction without that. So, let's try and see how it works.

I also think that we should put these macros into the `errs.h` module, as they directly relate to pushing errors onto the error stack. I think we can start all of their names with `TLOG_ERRS_RAISE`, as they would essentially do what raising exception does: exit a function with an error message'.

The most basic of them could be this:

    TLOG_ERRS_RAISEF("Failed to read %s", filename);

And more complicated could be:

    TLOG_ERRS_RAISECF(TLOG_GRC_FAILURE, "Failed to parse %s", filename)

or something like that. They would need to update the local `grc` variable (where needed), *perrs, and go to `cleanup` label.  See what you can come up with.

danghai:
Hi Nick, I write implementation about error handling macro in `err.h` module like you mentioned above. I do not write macro relating `pushf` yet. Please review this commit: 
https://github.com/danghai/tlog/commit/0d89853275183db6c7590338d0db59ad157b4294
 I have some problem with my master branch with your update's refactor code.  I do not know why it has 6 commit behind and 6 commit ahead.

danghai:
Actually I fix my master branch now. It is even with your Scribery:master now. But I do not know how to merge to fix the conflict in this branch. I mean I want to remove 6 redundant commit with only 1 my commit.

include/tlog/errs.h
```diff
@@ -141,4 +141,36 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
 extern int tlog_errs_print(FILE *stream, const struct tlog_errs *errs);
 
+#define TLOG_ERRS_RAISES(_perrs, _msg) \
+     do {                                       \
+         struct tlog_errs **__perrs = (_perrs); \
+         tlog_errs_pushs(__perrs, _msg);        \
+         goto cleanup;                          \
+     } while (0)
+
+#define TLOG_ERRS_RAISECS(_grc, _perrs, _msg) \
```
spbnick: I think we need to make all the macros use format strings (i.e. `tlog_errs_pushf`, for simplicity). Also, please name the format string argument `_fmt`, not `_msg`, since that's what it is, similarly to the way it is in `tlog_errs_pushf`

spbnick:
Macro argument names have underscore in front to avoid possible clashes with local variables, which don't usually have underscores in front.

spbnick:
That is, the argument names themselves shouldn't have a problem if they match a local name, but other variables declared inside could, and they should be prefixed. Arguments are prefixed for consistency. Frankly, I don't remember where exactly the convention came from, I just follow it, but the above explanation makes sense to me.

danghai:
I got it. 
I remove `_perrs` argument in macro. However I am little bit confused about using `struct tlog_err`. Sometime, I see you use variable name `*errs` ( For example: https://github.com/Scribery/tlog/blob/master/src/tlog/tlog-play.c#L45). Sometime, I see you use variable name `perrs` ( For example: https://github.com/Scribery/tlog/blob/master/lib/tlog/play.c#L310). Why don't you use the same name for all file, and then we can replace them with macro. 
I also document all the macros, similarly to other macros in the code.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+     do {                                       \
+         tlog_errs_pushs(perrs, _fmt);          \
+         goto cleanup;                          \
+     } while (0)
+
+/**
+ * Push an error to a stack, with message being a global return code
+ * description, update the local grc variable
+ * Push an error to a stack, with message as a constant string
+ * and then go to cleanup
```
spbnick: Please don't just write separate sentences about what this function is doing. Instead start it with something like: "Push two errors on the stack: ..."

spbnick: Also mention assigning `_grc` to `grc`.

danghai:
Hi Nick, I modified these change.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
  */
 extern int tlog_errs_print(FILE *stream, const struct tlog_errs *errs);
 
+/**
+ * Push an error to the perrs stack, with message as a constant string
```
spbnick: Since we're referring to the literal name "perrs" here, it would be better to quote it with double quotes.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ * and then go to cleanup.
+ *
+ * @param fmt   The message string to copy and put into the pushed error.
```
spbnick: The "fmt" name here is wrong, we're not passing a format string here, but just a message, so it should be "msg" or such instead. Also, please make sure that the leading underscore is present and argument names in the description and in the macro definition match.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+#define TLOG_ERRS_RAISES(_fmt) \
+     do {                                       \
+         tlog_errs_pushs(perrs, _fmt);          \
+         goto cleanup;                          \
+     } while (0)
+
+/**
+ * Push two errors to the perrs stack, with message being a global return code
+ * description, and with message as a constant string.
+ * Update the local grc variable by assigning _grc to grc,
```
spbnick: The macro doesn't care if "grc" is local or global. It should also be quoted. So something like "Assign _grc argument to "grc"" would work better.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ * and then go to cleanup.
+ *
+ * @param grc   The global return code, which description will be pushed to
+ *              the stack.
+ * @param fmt   The message string to copy and put into the pushed error.
```
spbnick: Same here with argument name.

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ */
+#define TLOG_ERRS_RAISES(_fmt) \
+     do {                                       \
+         tlog_errs_pushs(perrs, _fmt);          \
+         goto cleanup;                          \
+     } while (0)
+
+/**
+ * Push two errors to the perrs stack, with message being a global return code
+ * description, and with message as a constant string.
```
spbnick: This is good, but the sentence would be clearer this way: "first one with message being a global return code description, and the second with message as a constant string".

include/tlog/errs.h
```diff
@@ -141,4 +141,68 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ */
+#define TLOG_ERRS_RAISEF(_fmt, _args...) \
+     do {                                       \
+         tlog_errs_pushf(perrs, _fmt, ##_args); \
+         goto cleanup;                          \
+     } while (0)
+
+/**
+ * Push two errors to the perrs stack, with message being a global return code
+ * description, and with a sprintf-formatted message.
```
spbnick: Same here about clarity.

danghai:
I modified these change. Now, I know the `fmt` represents `format string`. :) I thought it represents all string before. What is the different between `format string` and `message`? Thanks

danghai:
One more issue: when I try to use the "TLOG_ERRS_RAISEF" and "TLOG_ERRS_RAISECF", it has `warning` in `make install`. But it still works.
```
play.c:241:33: warning: too many arguments for format [-Wformat-extra-args]
           TLOG_ERRS_RAISECF(grc,"TEST \n",str);
                                 ^
/home/danghai/tlog/include/tlog/errs.h:206:33: note: in definition of macro ‘TLOG_ERRS_RAISE
```
The output: 

```
play.c:tlog_play_create_json_reader:241
No such file or directory
TEST 

Failed creating log source

```

spbnick:
Since those error messages are presented to users, I don't think it would be good to make it more difficult to read by adding source code information. It might be a good idea to add that information in the debug build (i.e. without NDEBUG defined). However, that should be a separate PR.

include/tlog/errs.h
```diff
@@ -141,4 +141,70 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+/**
+ * Push an error to the "perrs" stack, with message as a constant string
+ * and then go to cleanup.
+ *
+ * @param _msg   The message string to copy and put into the pushed error.
+ *              Cannot be NULL.
+ */
+#define TLOG_ERRS_RAISES(_msg) \
+     do {                                                                 \
+         fprintf(stderr,"%s:%s:%d\n", __FILE__, __FUNCTION__, __LINE__);  \
```
spbnick: First, this change shouldn't be a part of this commit, or PR. Then, it goes against the idea of the "error stack", part which is to keep errors until a later time, when we can meaningfully output them. E.g. when the terminal we run under is in raw mode, the messages we print will not print nicely. Another part is that the caller decides when to output those error messages. With your change, the source code information will be printed immediately, no matter if the caller and the consumer of the error stack would even want to output it at all.

The source code information should be incorporated into the error stack itself, and attached to each error message, the item in the stack. It might be a good exercise to figure out how to best do that. Still, it shouldn't be a part of this PR.

include/tlog/errs.h
```diff
@@ -141,4 +141,70 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+         tlog_errs_pushs(perrs, _msg);                                    \
+         goto cleanup;                                                    \
+     } while (0)
+
+/**
+ * Push two errors to the "perrs" stack, first one with message being a global
+ * return code description, and the second with message as a constant string.
+ * Assign _grc argument to "grc, and then go to cleanup.
```
spbnick: Please add the missing double quote.

spbnick:
Regarding the warning you get trying to use it.

First of all, you should always configure your build with `--enable-debug` (i.e. run `./configure --enable-debug`), so that assertions are enabled and build warnings are promoted to errors. The `--enable-debug` option is implemented in `configure.ac`, take a look.

Second, the warning you get is due to gcc detecting that you supplied an argument (`str`) to your format string `"TEST \n"`, but the format string didn't have any format specifiers. The checking is enabled with special syntax in the declaration of `tlog_errs_pushf`, check it out. Also see https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute

danghai:
Thanks Nick, I rebased to the latest commit, remove the fprintf's, and fix the other issues. You can test it now.

include/tlog/errs.h
```diff
@@ -141,4 +141,66 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+         tlog_errs_pushc(perrs, grc);                                     \
+         tlog_errs_pushs(perrs, _msg);                                    \
+         goto cleanup;                                                    \
+     } while (0)
+
+/**
+ * Push an error to the "perrs" stack, with a sprintf-formatted message.
+ * and then go to cleanup.
+ *
+ * @param fmt   The message sprintf(3) format string. Cannot be NULL.
```
spbnick: Please make the parameter name here match the definition. I.e. prepend `fmt` with `_`.

include/tlog/errs.h
```diff
@@ -141,4 +141,66 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+     } while (0)
+
+/**
+ * Push two errors to the "perrs" stack, first one with message being a global
+ * return code description, and the second with a sprintf-formatted message.
+ * Assign _grc argument to "grc", and then go to cleanup.
+ *
+ * @param grc   The global return code, which description will be pushed to
+ *              the stack.
+ * @param fmt   The message sprintf(3) format string. Cannot be NULL.
```
spbnick: Same here.

danghai:
I have updated. There are some issues: 
1) In my opinion, I think I will try 3 functions in `src` including `tlog-play`, `tlog-rec`, `tlog-rec-session` first. After done without any issue, I will replace all error macro for the rest of the code in source. 

2) Sometime you update the `grc`, sometime not following `tlog_errs_pushs`. For example:
Not update `grc`: https://github.com/Scribery/tlog/blob/master/src/tlog/tlog-rec.c#L129
Update `grc`: https://github.com/Scribery/tlog/blob/master/lib/tlog/play.c#L92

I think I will add update `grc` in `TLOG_ERRS_RAISES(_msg)` too, It looks like keep the same value `grc` if you do not want to update `grc`. Similarly for `TLOG_ERRS_RAISEF`. 

3) There are some issues passing argument in `TLOG_ERRS_RAISEF` and `TLOG_ERRS_RAISECF`.

danghai:
Ah I have mistake in switching code. Please ignore the issue 2 and 3 at above comment. Please review this commit.

spbnick:
Looking at this, I think assigning to `grc` in those macros makes their use a bit too obscure and confusing. How about we remove that from the macros and assign to `grc` explicitly? Could you do that? Let's see how it looks.

danghai:
Sure, I will remove the grc from the macros and the code replacement could be: 
Ex: 

```c
         if (sigaction(tlog_play_exit_sig_list[i], NULL, &sa) == -1) {
            grc = TLOG_GRC_ERRNO;
            TLOG_ERRS_RAISECS("Failed to retrieve an exit "
                                              "signal action");
         }
```

Anyway in this commit, It still keeps the grc in these macro. I see it works normally. For example of code replacement:

```c
     grc = tlog_session_format_lock_path(&path, id);
     if (grc != TLOG_RC_OK) {
        TLOG_ERRS_RAISECS(grc, "Failed formatting lock file path");
     }
```

It has some confliction when I rebase this commit. I will fix it now.

danghai:
I updated switch the code to use error handling macro ( Except `*m4` file).

spbnick:
Alright, I think this looks better! Thank you, Hai. After looking at it for a little more, I think it would be now more natural to pass the `grc` to the macros explicitly as well, instead of taking it from the variable implicitly. Can you do that?

danghai:
Thank you Nick,
I pass the `grc` to the macros explicitly.

include/tlog/errs.h
```diff
@@ -141,4 +141,64 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ * And then go to cleanup.
+ *
+ * @param _grc  The global return code, which description will be pushed to
+ *              the stack.
+ * @param _msg  The message string to copy and put into the pushed error.
+ *              Cannot be NULL.
+ */
+#define TLOG_ERRS_RAISECS(_grc, _msg) \
+     do {                                                                 \
+         tlog_errs_pushc(perrs, grc);                                     \
```
spbnick: This should use `_grc` as well.

include/tlog/errs.h
```diff
@@ -141,4 +141,64 @@ extern void tlog_errs_pushf(struct tlog_errs **perrs, const char *fmt, ...)
...
+ * And then go to cleanup.
+ *
+ * @param _grc  The global return code, which description will be pushed to
+ *              the stack.
+ * @param _fmt  The message sprintf(3) format string. Cannot be NULL.
+ * @param ...   The message format arguments.
+ */
+#define TLOG_ERRS_RAISECF(_grc, _fmt, _args...) \
+     do {                                                                 \
+         tlog_errs_pushc(perrs, grc);                                     \
```
spbnick: Same here.

lib/tlog/session.c
```diff
@@ -99,45 +99,36 @@ tlog_session_format_lock_path(char **ppath, unsigned int id)
...
+        if (seteuid(_orig_euid) < 0) {                                   \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed restoring EUID");             \
+        }                                                                \
+                                                                         \
+        /* Restore EGID */                                               \
+        if (setegid(_orig_egid) < 0) {                                   \
+            grc = TLOG_GRC_ERRNO;                                        \
+            TLOG_ERRS_RAISECS(grc, "Failed restoring EGID");             \
+        }                                                                \
```
spbnick: I think this is wrong. It's not using the __perrs pointer. We should either leave it as is, or remove the `_perrs` argument from the macro. I think the latter might be better. What do you think?

danghai:
I think removing the `_perrs` is better way ( It looks like the same these macro). I updated the PR.

spbnick:
This looks just great, merged. Thank you very much, Hai!

Issue #150, opened 2017-12-15
Do not exit tlog-play with "-g end -p" options


spbnick:
When both `--goto=end` and `--paused` options are specified to tlog-play, it exits after going to the end. Instead it should stay on pause.

danghai:
Hi Nick, I add the `paused` flag in `if (json_object_object_get_ex(conf, "goto", &obj)) {` condition. It will stay on pause instead exiting after going to the end.  Please review the commit: 
if (json_object_object_get_ex(conf, "goto", &obj)) {

spbnick:
Hi Hai, thank you for trying to fix this. However, we still need "goto" to work here, we only need to not exit and stay on pause if we reach the end of recording. See the code which reacts to end-of-recording received from the source and checks for the "goto" being active.

danghai:
For clear, `only` both flag `--goto=end` and `paused` are set at same time (ex: `tlog-play -i tlog.log --goto=END -p`), it should stay on pause instead of exiting? And it normally exits with other options ( Ex: `tlog-play -i tlog.log`: It will play and exit normally after reaching the end of recording). Please review this commit: https://github.com/danghai/tlog/commit/8a45927513fa0650235891cfae3edb19b7f496d4
I handle this by declare one more variable name `init_paused` ( because paused is modified by `space` ), if init_paused is set, it generates `paused` flag.

spbnick:
"goto" works with "paused": it goes to the specified location and stays on pause. Try executing `tlog-play` with `--goto=10` (or some such) and `--paused` and see. It should behave the same for `--goto=end`. I think it should be possible to implement that without introducing a new variable.

spbnick:
Fixed in 1671124b83b67360c392805999112f564128b422. Thanks, Hai!

PR #155, opened 2017-12-23
Do not exit tlog-play with "-g end -p" options

danghai:
Hi Nick, pull request for this issue: https://github.com/Scribery/tlog/issues/150. 
I handle the issue 150 without using more variable. Executing `--goto=10 -p` is the same `--goto=END -p`. It does not exit after going to end. Instead it should stay on pause.
I realize that I do not follow your style coding ( open braces) in the first commit. I modified it too.  
Sorry, I do not know why my commit go along with your commit's refactor

### DIFF
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -927,6 +927,8 @@ tlog_play_run(struct tlog_errs **perrs, int *psignal)
                 if (tlog_play_follow) {
                     read_wait = (struct timespec){POLL_PERIOD, 0};
                     continue;
+                } else if (tlog_play_paused) {
+                    continue;
                 } else {
                     break;
                 }


danghai:
I think I can fix the conflict between master and my branch by git rebase. I reopen the issue because the old one cannot reopen. Please review the commit for this issue

lib/tlog/play.c
```diff
@@ -924,9 +924,17 @@ tlog_play_run(struct tlog_errs **perrs, int *psignal)
             /* If hit the end of stream */
             if (tlog_pkt_is_void(&pkt)) {
                 tlog_play_goto_active = false;
-                if (tlog_play_follow) {
-                    read_wait = (struct timespec){POLL_PERIOD, 0};
-                    continue;
+                if (tlog_play_paused) {
+                    do {
+                        rc = clock_nanosleep(CLOCK_MONOTONIC, 0,
+                                             &tlog_timespec_max, NULL);
```
spbnick: We shouldn't be copying pausing logic here. See if you can avoid that. It would be best if pausing would still be handled by the same (maybe expanded) code.

danghai:
Hi Nick, Could you explain why I cannot use copy pausing logic here. I see it can fix this issue without bugs? You want me expand the code right here: https://github.com/Scribery/tlog/blob/master/lib/tlog/play.c#L771 to handle this issue? 
Another question is that what is the different between 'p' and 'space' control key. I mean it has the same function (pausing), but the implementation is different. Thanks

spbnick:
I don't want to copy pausing logic, because that increases the chance that the copies will diverge over time producing different pausing behavior in different cases. For the same reason you generally avoid copy-pasting any data or code (with exceptions).

No, I don't think you need to touch the code reacting to user input and changing the paused state. What I meant is that paused state handling should stay in the same place, namely where it says "Handle pausing, unless ignoring timing".

The implementation for "p" and "space" keys should be the same, if not, we need to fix that. Could you elaborate? If instead you're asking why we have two keys for the same function, then it's because we couldn't support "space" in the Cockpit UI (as it's special to browsers) and added another key ("p"), and made tlog-play mirror that change. That was back when we still used tlog-play in a terminal instead of implementing playback in JavaScript.

danghai:
I create another pull request to fix the open brace. 
I think it is better to add `if condition` with continue when tlog hit the end of stream. If 'tlog_play_paused' is set, it will paused at end of stream. I do not need to add 'paused' logic like previous commit. It is actually better way to fix that. Please review this commit. 
I am asking why we have two keys for the same function. Your answer makes sense to me. Because I do not know about Cockpit UI, but using "p" key control instead of "space" (special ) makes sense to me.

spbnick:
Hey, this is great, merged! Thanks, Hai!

PR #157, opened 2017-12-27
Implement tlog-play "to" function

danghai:
As we know, the function `"goto"` is the fast-forward to STRING time. Tlog-play will play recording from the STRING time that we set until to the end. I implemented the `"to"` function that Tlog-play will `STOP` at STRING time. I think it is convenient because : 
1) Sometimes, users do not want to play recording until the END. They want to stop at specific time. For example: 
`tlog-play -i clock.log --to=10` .
 It will play recording from `START` to `10` and exit. 

2) For cooperate with `"goto"` function. Usually, the recording is too long, and the users want to focus on the specific range time. They can fast-forward to specific time and then play the recording until hitting the specific time. For instance: 
`tlog-play -i clock.log --goto=150 --to=200`
It will play recording from 150 - 200 ( the specific range time)

I modified `play_conf_schema.m4` to generate the play_conf_cmd.c with one more option "to". In `play.c` I declare 2 variable ( one variable for storing the timestamp and one for flag). What do you think?

### DIFF
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -338,6 +338,10 @@ struct timespec tlog_play_speed;
 bool tlog_play_goto_active;
 /** Timestamp the "goto" function should go to */
 struct timespec tlog_play_goto_ts;
+/** True if "to" function is active */
+bool tlog_play_to_active;
+/** Timestamp the "to" function should go to */
+struct timespec tlog_play_to_ts;
 /** True if "skip" function is active */
 bool tlog_play_skip;
 /** True if playback is paused */
@@ -459,6 +463,7 @@ tlog_play_init(struct tlog_errs **perrs,
     tlog_play_speed.tv_sec = 1;
     tlog_play_speed.tv_nsec = 0;
     tlog_play_goto_active = false;
+    tlog_play_to_active = false;
     tlog_play_skip = false;
     tlog_play_paused = false;
     tlog_play_persist = false;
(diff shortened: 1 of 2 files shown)


danghai:
3. User can set `STOP` while playing recording like fast-forward by `'T` character

spbnick:
This feature seems nice, however, I don't see it being particularly useful, and unless we get requests from actual users for something like this, I would prefer not to implement it to avoid slowing us down during refactoring and implementing other features.

If you'd like, continue developing this, and I can give you some comments and feedback, but I won't merge it for now.

General comments so far: the "to" name for the option is confusing - too similar to "goto". Something like "--end-at=" would be better. I think tlog-play shouldn't exit when reaching the specified location, if "--follow" was specified, or if, e.g. the playback was fast-forwarded to the end location while on pause. I.e. it should behave exactly as if the end of the stream was reached at the specified location and nothing else was coming.

Good job figuring out the way to add the option, BTW!

danghai:
Thank you for your feedback. It is just my idea when I want to learn how to write more option to your tlog. It is very nice to get your comments and feedback to continue developing. However, as you said above, when new feature is added, a lot of issues, problem, bugs will come after. It is better for me to focus on the current issue rather than continue developing it. You can close this pull request, but if you want to develop this feature in future, I am willing to work on it.

spbnick:
Sure. Thank you, Hai. I'll close this for now, and we can find and reopen this later, if needed.

PR #161, opened 2018-01-18
Fix up brace use and double semicolon

danghai:
The pull request for fixing up brace use and removing double semicolon


spbnick:
Thanks Hai, merged!

PR #172, opened 2018-01-28
Refactor tlog-play source creation

### LINKED ISSUE #141: Refactor tlog-play source creation
Split tlog-play source creation into functions similarly to what was done to sink creation in lib/rec.c

danghai:
Fixes #141 
I already split `tlog_play_create_json_reader` into 3 functions per reader type regarding to `es`, `journal`, and `file`. The name functions include: 
a. tlog_play_create_es_json_reader
b. tlog_play_create_journal_json_reader
c. tlog_play_create_file_json_reader

Issue: 
1. I put `es/journal/file` reader conf container into each function instead of at `tlog_play_create_json_reader`. It is different `rec.c` (reader conf container is at tlog_rec_create_json_writer).
2. I already test tlog-play with `file`, and it works; Unfortunately, I am not sure about test tlog-play with `es` and `journal`.
3. Is it better to use `switch-case` instead of using `if-else` condition at `tlog_play_create_json_reader`, as well as `tlog_rec_create_json_writer` ?

Please review this commit: 
https://github.com/danghai/tlog/commit/d941e98c27bc7bb291eea1409331a66f51901440
file:
https://github.com/danghai/tlog/blob/tlog_issue141/lib/tlog/play.c

### DIFF
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -65,6 +65,235 @@ tlog_play_io_sighandler(int signum)
 }
 
 /**
+ * Create an Elasticsearch JSON message reader according to configuration.
+ *
+ * @param perrs         Location for the error stack. Can be NULL.
+ * @param preader       Location for the created reader pointer. Not modified
+ *                      in case of error.
+ * @param conf          Configuration JSON object.
+ *
+ * @return Global return code.
+ */
+static tlog_grc
+tlog_play_create_es_json_reader(struct tlog_errs **perrs,
+                                struct tlog_json_reader **preader,
+                                struct json_object *conf)
+{
+    tlog_grc grc;
+    const char *baseurl;
+    const char *query;
+    struct tlog_json_reader *reader = NULL;
+    struct json_object *obj;
+
+    assert(preader != NULL);
+    assert(conf != NULL);
+
+    /* Get the base URL */
+    if (!json_object_object_get_ex(conf, "baseurl", &obj)) {
+        tlog_errs_pushs(perrs, "Elasticsearch base URL is not specified");
+        grc = TLOG_RC_FAILURE;
+        goto cleanup;
+    }
+    baseurl = json_object_get_string(obj);
+
+    /* Check base URL validity */
+    if (!tlog_es_json_reader_base_url_is_valid(baseurl)) {
+        tlog_errs_pushf(perrs,
+                        "Invalid Elasticsearch base URL: %s", baseurl);
+        grc = TLOG_RC_FAILURE;
+        goto cleanup;
+    }
+
+    /* Get the query */
+    if (!json_object_object_get_ex(conf, "query", &obj)) {
+        tlog_errs_pushs(perrs, "Elasticsearch query is not specified");
+        grc = TLOG_RC_FAILURE;
+        goto cleanup;
+    }
+    query = json_object_get_string(obj);
+
+    /* Create the reader */
+    grc = tlog_es_json_reader_create(&reader, baseurl, query, 10);
+    if (grc != TLOG_RC_OK) {
+        tlog_errs_pushc(perrs, grc);
+        tlog_errs_pushs(perrs, "Failed creating the Elasticsearch reader");
+        goto cleanup;
+    }
+
+    *preader = reader;
+    reader = NULL;
+    grc = TLOG_RC_OK;
+
+cleanup:
+    tlog_json_reader_destroy(reader);
+    return grc;
+}
+
+#ifdef TLOG_JOURNAL_ENABLED
+/**
+ * Create a journal JSON message reader according to configuration.
+ *
+ * @param perrs         Location for the error stack. Can be NULL.
+ * @param preader       Location for the created reader pointer. Not modified
+ *                      in case of error.
+ * @param conf          Configuration JSON object.
+ *
+ * @return Global return code.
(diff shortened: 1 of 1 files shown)


spbnick:
Thank you, Hai.

Please leave comments and questions concerning a pull request in that pull request's comments. Otherwise it's hard to find what they talk about, and answering them in the issue would lead to further confusion.

Also, instead of putting the issue number into the pull request description. Put the issue number into the commit message and format it according to "[Closing issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/)". E.g. in this case you can write in commit message:

    Fixes #141 

Then when we merge the commit, the issue would get automatically closed.

Next, regarding the code:
* It is better to keep extracting the configuration container outside the function that works on that container. This keeps the function independent of where the container is located. This way you can equally well use the function for extracting data from a single container, from two different containers with different names (by calling it twice, with different parameters), or from members of 100-element array (by calling it 100 times). Not that we need that at this moment, but it might come in handy and, in general, makes the function less coupled to the rest of the code, which is often good. It is however, good, that you're applying your own reasoning and experiment with the design, not simply copying what's already done.
* Regarding testing tlog with journal, see the updated README.md, it is quite easy.
* Regarding testing with Elasticsearch, it is harder, but might be good for you to try. I would recommend creating a VM (e.g. with virt-manager, installing Elasticsearch and Kibana there according to documentation at https://www.elastic.co and setting up log forwarding and playback as described in README.md. I will leave that up to you, however, and can test that myself.
* Regarding switch-case, I'd like to leave it up to you to try, and figure it out.
* A specific comment about the code for now, is to please make sure your indenting is correct, and there are no unnecessary, or missing blank lines.

Thank you!

lib/tlog/play.c
```diff
@@ -64,6 +64,234 @@ tlog_play_io_sighandler(int signum)
...
+        str_list = calloc(json_object_array_length(obj) + 1,
+                          sizeof(*str_list));
+        if (str_list == NULL) {
+            grc = TLOG_GRC_ERRNO;
+            tlog_errs_pushc(perrs, grc);
+            tlog_errs_pushs(perrs,
+                            "Failed allocating systemd journal match list");
+            goto cleanup;
+        }
+        for (i = 0; (int)i < (int)json_object_array_length(obj); i++) {
```
spbnick: Indentation is off by one space below this line...

lib/tlog/play.c
```diff
@@ -64,6 +64,234 @@ tlog_play_io_sighandler(int signum)
...
+                 grc = TLOG_RC_FAILURE;
+                 tlog_errs_pushf(
+                     perrs,
+                     "Systemd journal match symbol #%zu \"%s\" "
+                     "is invalid",
+                     i + 1, str_list[i]);
+                 goto cleanup;
+             }
+         }
+     }
```
spbnick: ... and until and including this line. Please fix.

lib/tlog/play.c
```diff
@@ -64,6 +64,234 @@ tlog_play_io_sighandler(int signum)
...
+ * @param preader       Location for the created reader pointer. Not modified
+ *                      in case of error.
+ * @param conf          Configuration JSON object.
+ *
+ * @return Global return code.
+ */
+ static tlog_grc
+ tlog_play_create_file_json_reader(struct tlog_errs **perrs,
+                                   struct tlog_json_reader **preader,
+                                   struct json_object *conf)
```
spbnick: This shouldn't be indented.

lib/tlog/play.c
```diff
@@ -64,6 +64,234 @@ tlog_play_io_sighandler(int signum)
...
+         tlog_errs_pushs(perrs, "Failed creating file reader");
+         goto cleanup;
+     }
+     fd = -1;
+
+     *preader = reader;
+     reader = NULL;
+     grc = TLOG_RC_OK;
+
+cleanup:
```
spbnick: The body of this function above this line has one extra space of indent. Please fix.

danghai:
I fixed the indenting issues.

lib/tlog/play.c
```diff
@@ -80,172 +309,60 @@ tlog_play_create_json_reader(struct tlog_errs **perrs,
...
-                            perrs,
-                            "Systemd journal match symbol #%zu \"%s\" "
-                            "is invalid",
-                            i + 1, str_list[i]);
-                        goto cleanup;
-                    }
-                }
-            }
+        if (!json_object_object_get_ex(conf, str, &reader_conf)) {
+            tlog_errs_pushs(perrs, "Journal reader parameters are not specified");
```
spbnick: This breaks running `tlog-play` without journal reader parameters (we need integration tests ASAP to catch these). I.e. as just `tlog-play -r journal`. This normally is not useful, but in general should be supported. You can fix this by allowing passing `NULL` `conf` to the reader-creating function. All the `json_*` functions would then act as if the node you're retrieving doesn't exist, and it should work OK. Also, while you're updating and moving around these lines, please make sure that they are within 78 chars long. Thanks!

danghai:
Thanks Nick, I have updated it !

spbnick:
Please rebase this on the latest master and we'll likely merge it.

danghai:
I rebased this commit on the latest master.

spbnick:
Merged, thank you, Hai!

PR #176, opened 2018-02-05
Add RPM build to CI runs

### LINKED ISSUE #168: Add RPM build to CI runs

danghai:
Fixes #168 
Pull request for add RPM build to CI runs. 
In the `.travis.yml`, why do you want to use the `make distcheck`, But in the `README`, you use 
`./configure --prefix=/usr --sysconfdir=/etc && make dist` ?
When I build RPM in local machine: `rpmbuild -ta tlog-5.tar.gz` ( Following README), It produces the error 
```
error: Failed build dependencies:
	curl-devel is needed by tlog-5-1.x86_64
	json-c-devel is needed by tlog-5-1.x86_64
	m4 is needed by tlog-5-1.x86_64
	systemd-devel is needed by tlog-5-1.x86_64
	systemd-units is needed by tlog-5-1.x86_64
```

However, when I build with command the same in `travis` (rpmbuild --define "_tmpfilesdir /usr/lib/tmpfiles.d" -ta --nodeps tlog-5.tar.gz), then it works. Could you explain why? Thanks


danghai:
actually, it is fail when I use `make distcheck`. However, it work with `make dist`. What about introduction deployment packagecloud: https://docs.travis-ci.com/user/deployment/packagecloud. 
When I read information about rpm in `Travis-CI`, they mention about `docker`. I am not sure about it.

spbnick:
Right, `make distcheck` finds the coverage data files which apparently `scan-build`-controlled build produces. We can either go without `scan-build` with `make distcheck`, or add cleanup of `.gcno` files to Makefile.am's. See https://www.gnu.org/software/automake/manual/automake.html#Extending for the latter.

I added issue #177 to keep the RPM deployment in mind, but for now let's go without it.

Can you give a link to Travis-CI talking about Docker?

spbnick:
Regarding the question about failed build dependencies, first, what distro are you running on your local machine?

danghai:
I am using Ubuntu 17.10.1. 
I read something github link following link: 
https://djw8605.github.io/2016/05/03/building-centos-packages-on-travisci/
https://gist.github.com/abn/daf262e7e454509df1429c87068923d1
https://hub.docker.com/r/mgruener/rpmbuild/
https://github.com/packpack/packpack
https://twindb.com/building-rpm-on-travis-ci-in-docker-containers/

However, I did not apply anything from there. Sometime, I do not understand what they do. I read your instruction about how to build `rpm`, and then read something in Travis website. 
Do I need to fix anything in `.travis.yml`

spbnick:
If you're running `rpmbuild` on Ubuntu, that means you don't actually have anything in the `rpm` database, since Ubuntu is using `dpkg` (as do all Debian-derived distros). For that reason your command line in `.travis.yml` is passing the `--nodeps` option to `rpmbuild`. If you do the same on your local machine (after ensuring dependencies are actually installed), it will work too.

Ah, I see, they're using Docker to get the exact OS environment they're building the RPM for. However, we are not building the RPM here for a release, but only to check that it can build in principle. Building it directly on Ubuntu would be a good enough test for now. Certainly better than not building it. Adding Docker to it would increase complexity and the length of the CI run.

While there are plenty of build systems out there, the autotools (the ./configure && make approach) are still the most standard, most supported, most understood, and most portable one. Also, I know it quite well :) Yeah, it can be slow and hard to figure out, but let's keep to it for now.

.travis.yml
```diff
@@ -16,13 +16,17 @@ before_install:
...
+- sudo apt-get install -y rpm
 - gem install coveralls-lcov
+- mkdir -p ~/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
 before_script:
 - autoreconf -i -f
 - $CLANG_SCAN_BUILD ./configure --disable-silent-rules $CONFIGURE_ARGS
 script:
 - $CLANG_SCAN_BUILD make
 - $CLANG_SCAN_BUILD make check
+- make dist
```
spbnick: This still needs to be `make distcheck`.

.travis.yml
```diff
@@ -16,13 +16,17 @@ before_install:
...
+- rpmbuild --define "_tmpfilesdir /usr/lib/tmpfiles.d" -ta --nodeps tlog-5.tar.gz
```
spbnick: This should not rely on a particular tlog version to work.

danghai:
Hi Nick, 
Thanks your the article you gave me about `automake`. It helps me a lot. According to https://www.gnu.org/software/automake/manual/automake.html#Preparing-Distributions, I think the reasons why `make distcheck` is fail are `VPATH` and `distcleancheck`. I do not know how to run ‘make distcleancheck’ at the end of the VPATH build. So, I disable `distcleancheck` in `automake` to make sure we can build `make distcheck`. 

I am not sure about you mention 
`add cleanup of .gcno files to Makefile.am's. See https://www.gnu.org/software/automake/manual/automake.html#Extending for the latter.` Could you elaborate? Anyway, I am happy to know little about coverall file. Sometime, I wonder  `coverage` (Ex: 37% now) evaluates based on what.

Makefile.am
```diff
@@ -23,3 +23,7 @@ ACLOCAL_AMFLAGS = -Im4/autotools
 SUBDIRS = m4 include lib src doc man
 dist_doc_DATA = README.md
 dist_noinst_DATA = tlog.spec
+
+# disable distcleancheck in order to make distcheck.
+distcleancheck:
+        @:
```
spbnick: We need this check, so please don't remove it, but good job figuring out how it works!

.travis.yml
```diff
@@ -16,13 +16,18 @@ before_install:
...
 - gem install coveralls-lcov
+- mkdir -p ~/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
 before_script:
 - autoreconf -i -f
 - $CLANG_SCAN_BUILD ./configure --disable-silent-rules $CONFIGURE_ARGS
 script:
+- ./configure --prefix=/usr
 - $CLANG_SCAN_BUILD make
 - $CLANG_SCAN_BUILD make check
+- make distcheck
```
spbnick: Did I get it right, that this still fails, even without `$CLANG_SCAN_BUILD`, because of the `.gcno` files? If so, then I was mistaken, and it's the coveralls, who asks gcc to create those files. A good fix would be to add removal of `.gcno` files from all directories where object files are created, to the `clean` target. I guess that can be done by adding the `clean-local` target to the affected Makefile.am`s (with a command to remove them) as described in https://www.gnu.org/software/automake/manual/automake.html#Extending

danghai:
Yeah, it still fails, even without `$CLAN_SCAN_BUILD`. I added removal of `.gcno` files. However, it is still wrong with `discleancheck` because of `VPATH` build.

Makefile.am
```diff
@@ -23,3 +23,7 @@ ACLOCAL_AMFLAGS = -Im4/autotools
 SUBDIRS = m4 include lib src doc man
 dist_doc_DATA = README.md
 dist_noinst_DATA = tlog.spec
+# Remove *.gcno files from all directories.
+clean-local:
+	-rm -rf *.gcno
```
spbnick: This is not how shell and rm works. It won't do what I think you tried to do here. You need to understand how shell does expansion. See https://www.gnu.org/software/bash/manual/html_node/Shell-Expansions.html and ping me with questions if needed.

Then, this Makefile.am is only responsible for the root directory directly. Each subdirectory has its own Makefile.am's to care for it. You need to add removal of .gcno files to Makefile.am of each directory, which has .gcno files produced. I.e. each directory which has .c files.

.travis.yml
```diff
@@ -16,13 +16,18 @@ before_install:
...
 - sudo apt-get install -y clang
 - sudo apt-get install -y lcov
+- sudo apt-get install -y rpm
 - gem install coveralls-lcov
+- mkdir -p ~/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
 before_script:
 - autoreconf -i -f
 - $CLANG_SCAN_BUILD ./configure --disable-silent-rules $CONFIGURE_ARGS
 script:
+- ./configure --prefix=/usr
```
spbnick: Hmm, why did you add this here?

danghai:
Hi Nick, 
I forget delete the line `./configure --prefix=/usr`. The reason why I put this line because, when I read the DISTCHECK_CONFIGURE_FLAGS at: https://www.gnu.org/software/automake/manual/automake.html#Checking-the-Distribution. 
I think I need to configure before `make distcheck`.
I update the delete *.gcno file in Makefile.am. It will find all file `*.gcno` in top level of the current build tree and delete all

danghai:
I added remove the *gcda file. However, it has the error: 
```
ERROR: files left in build directory after distclean:
./clang_output_K8uA66
./clang_output_J297dQ
```
I am not sure about it

Makefile.am
```diff
@@ -23,3 +23,7 @@ ACLOCAL_AMFLAGS = -Im4/autotools
 SUBDIRS = m4 include lib src doc man
 dist_doc_DATA = README.md
 dist_noinst_DATA = tlog.spec
+# Remove *.gcno files from all directories.
```
spbnick: This comment doesn't match what's being done, and is not really helpful. It's better to explain what those files are and where they come from. Please fix.

Makefile.am
```diff
@@ -23,3 +23,7 @@ ACLOCAL_AMFLAGS = -Im4/autotools
...
+clean-local:
+	find $(top_builddir) -name '*.gcno' -delete
+	find $(top_builddir) -name '*.gcda' -delete
```
spbnick: This is not the best approach and a proper one would be putting `clean-local` target into every Makefile.am which builds .c files instead, but perhaps this will do. Please quote `$(top_builddir)` with single quotes.

Lastly, this can be combined into a single command like this:

    find '$(top_builddir)' -name '*.gcno' -or -name '*.gcda' -delete

danghai:
Hi Nick, It has some issues:
1) I put `$CLANG_SCAN_BUILD` to `make distcheck`. It has bugs on Travis-CI: 
```
scan-build: 8 bugs found.
scan-build: Run 'scan-view /tmp/scan-build-2018-02-20-191800-16354-1' to examine bug reports.
scan-build: The analyzer encountered problems on some source files.
scan-build: Preprocessed versions of these sources were deposited in '/tmp/scan-build-2018-02-20-191800-16354-1/failures'.
scan-build: Please consider submitting a bug report using these files:
scan-build:   http://clang-analyzer.llvm.org/filing_bugs.html
```
2) Could you recheck with this command: ( It does not work)
```
find '$(top_builddir)' -name '*.gcno' -or -name '*.gcda' -delete
```

3) I read something relating to `clean` in automake: 
https://www.gnu.org/software/automake/manual/html_node/Clean.html
They mention: 
```
Generally the files that can be cleaned are determined automatically by Automake. Of course, Automake also recognizes some variables that can be defined to specify additional files to clean. These variables are MOSTLYCLEANFILES, CLEANFILES, DISTCLEANFILES, and MAINTAINERCLEANFILES.
```
I try to use `CLEANFILES` without these command ( `find '$(top_builddir)' -name '*.gcno' -delete` , ...) Because I think it is cleaned automatically by Automake. However, it does not work. 

4) It works when I put command removing `clang_output_*` , and only `make distcheck` ( without `$CLANG`)

spbnick:
> Hi Nick, It has some issues:
> 1) I put `$CLANG_SCAN_BUILD` to `make distcheck`. It has bugs on Travis-CI: 
> https://travis-ci.org/danghai/tlog/jobs/343982292
> ```
> scan-build: 8 bugs found.
> scan-build: Run 'scan-view /tmp/scan-build-2018-02-20-191800-16354-1' to examine bug reports.
> scan-build: The analyzer encountered problems on some source files.
> scan-build: Preprocessed versions of these sources were deposited in '/tmp/scan-build-2018-02-20-191800-16354-1/failures'.
> scan-build: Please consider submitting a bug report using these files:
> scan-build:   http://clang-analyzer.llvm.org/filing_bugs.html
> ```

Alright, that seems to be Clang finding issues with some programs configure is
compiling to test some features, which doesn't concern us. However, that means
we cannot use `$CLANG_SCAN_BUILD` with `make distcheck`. So let's remove it.

> 2) Could you recheck with this command: ( It does not work)
> ```
> find '$(top_builddir)' -name '*.gcno' -or -name '*.gcda' -delete
> ```

Right, I wasn't aware `-delete` worked like that. The proper command would be:

    find '$(top_builddir)' -name '*.gcno' -delete -or -name '*.gcda' -delete

> 3) I read something relating to `clean` in automake: 
> https://www.gnu.org/software/automake/manual/html_node/Clean.html
> They mention: 
> ```
> Generally the files that can be cleaned are determined automatically by Automake. 
> Of course, Automake also recognizes some variables that can be defined to specify 
> additional files to clean. These variables are MOSTLYCLEANFILES, CLEANFILES, 
> DISTCLEANFILES, and MAINTAINERCLEANFILES.
> ```
> I try to use `CLEANFILES` without these command ( `find '$(top_builddir)' -name '*.gcno' -delete` , ...) Because I think it is cleaned automatically by Automake. However, it does not work. 

If you'd like, I can take a look at what you tried, but in general we won't be
able to use those as easily as the `clean-local` target, as we have to find
those files.

> 4) It works when I put command removing `clang_output_*` , and only `make distcheck` ( without `$CLANG`) 

Ah good. Let's do that then.

danghai:
I have updated it and rebased with latest commit.

Makefile.am
```diff
@@ -23,3 +23,8 @@ ACLOCAL_AMFLAGS = -Im4/autotools
 SUBDIRS = m4 include lib src doc man
 dist_doc_DATA = README.md
 dist_noinst_DATA = tlog.spec
+
+# Remove coverage data files and clang results generated by CI runs
+clean-local:
+	find '$(top_builddir)' -name '*.gcno' -delete -or -name '*.gcda' -delete
+	find '$(top_builddir)' -name 'clang_output_*' -delete
```
spbnick: Ah, wait, we don't need to look for 'clang_output_*' files. They're put into the source root directory. A simple "rm" will do. Please amend. Otherwise this looks fine.

spbnick:
Thank you, Hai, merged!

PR #180, opened 2018-02-13
Fix logging output terminates with Operation not supported

### LINKED ISSUE #171: Logging /dev/urandom output terminates with "Operation not supported"
When `tlog-rec-session` is used to record the output of `cat /dev/urandom` to Journal, it quickly terminates with the following output:

    Operation not supported
    Failed logging terminal data
    Failed transferring TTY data

danghai:
Fixes #171 
Pull request to fix logging output terminates with `Operation not supported`. 
I see you use the hard-code `CLOCK_MONOTONIC` for `clock_nanosleep`. However, in `rl_json_writer.c`, you do not use hard-code.  `rl_json_writer->clock_id` may not initialize `CLOCK_MONOTONIC`. So, I try to use hard-code value. It works normal now.

### DIFF
--- a/lib/tlog/rec.c
+++ b/lib/tlog/rec.c
@@ -585,7 +585,6 @@ cleanup:
  * @param id            ID of the recording being created.
  * @param username      The name of the user being recorded.
  * @param session_id    The ID of the audit session being recorded.
- * @param clock_id      The clock to use for rate-limiting.
  *
  * @return Global return code.
  */
@@ -595,8 +594,7 @@ tlog_rec_create_json_writer(struct tlog_errs **perrs,
                             struct json_object *conf,
                             const char *id,
                             const char *username,
-                            unsigned int session_id,
-                            clockid_t clock_id)
+                            unsigned int session_id)
 {
     tlog_grc grc;
     struct json_object *obj;
@@ -681,7 +679,7 @@ tlog_rec_create_json_writer(struct tlog_errs **perrs,
 
     /* Create rate-limiting writer */
     grc = tlog_rec_create_rl_json_writer(perrs, &writer, writer_conf,
-                                         clock_id);
+                                         CLOCK_MONOTONIC);
     if (grc != TLOG_RC_OK) {
         goto cleanup;
     }
@@ -702,7 +700,6 @@ cleanup:
  * @param psink         Location for the created sink pointer.
  * @param conf          Configuration JSON object.
  * @param session_id    The ID of the session being recorded.
- * @param clock_id      The clock to use for rate-limiting.
  *
  * @return Global return code.
  */
@@ -710,8 +707,7 @@ static tlog_grc
 tlog_rec_create_log_sink(struct tlog_errs **perrs,
                          struct tlog_sink **psink,
                          struct json_object *conf,
-                         unsigned int session_id,
-                         clockid_t clock_id)
+                         unsigned int session_id)
 {
     tlog_grc grc;
     int64_t num;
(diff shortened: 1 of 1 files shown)


spbnick:
Well, that would simply be throwing away the whole logic of passing clock_id through to rl_json_writer. So the first question you need to ask is why it was put there in the first place, instead of just "shorting it out". That leaves unused code around, which is never good. It would be confusing and would make the code "rot" - get broken over time, as it would be unused.

Then, if it is really not needed, *delete* that machinery, and then switch to CLOCK_MONOTONIC. If it's used, then the proper fix would be to arrange for this machinery to use CLOCK_MONOTONIC instead, not just plug it in in the middle.

Another good question for you to ask would be what other value beside CLOCK_MONOTONIC is used, and why it is used.

danghai:
Hi Nick, thanks your feedback. After reading more, I think one of the main reason's  is using `CLOCK_MONOTONIC_COARSE` ( for faster) in `/lib/rec.c`. However I am still not sure why `clock_nanosleep` cannot use the `CLOCK_MONOTONIC_COARSE`. 
For the reason why it still works in Debian, but it does not work in Fedora. When I run under platform Debian, the `clock_id` stores `1` ( CLOCK_MONOTONIC) --> It works normally. However, when I run under platform Fedora, the `clock_id` stores `6` (CLOCK_MONOTONIC_COARSE). 
So, I change `ifndef` in rec.c to make them use CLOCK_MONOTONIC ( You can see it in commit). I have a question : how you can run the code under defined/not define "CLOCK_MONOTONIC_COARSE" by command? I mean how you can pick the clock `CLOCK_MONOTONIC_COARSE`, as well as `CLOCK_MONOTONIC` to run by command?

I try to figure out how I can use `CLOCK_MONOTONIC_COARSE` with `clock_nanosleep` because you recommend use it for faster.

spbnick:
Regarding why CLOCK_MONOTONIC_COARSE can't be used with clock_nanosleep, a good place to start reading would be https://github.com/torvalds/linux/blob/61f14c015f5be9151ba25e638d349f4d40cb7cd4/kernel/time/posix-timers.c#L1204

The ultimate reason, I guess, is the implementation of CLOCK_MONOTONIC_COARSE, which doesn't seem to be implemented as a real timer, but just a "timer variable" updated periodically (that's why it's "coarse" and is faster). I.e. it cannot be used to "wake up" a process when certain time is reached.

If you read the code in `lib/tlog/rec.c` you can see how either `CLOCK_MONOTONIC_COARSE`, or `CLOCK_MONOTONIC` is chosen. The check for `CLOCK_MONOTONIC_COARSE` being defined is there just for older systems, which didn't have it at all.

> I have a question : how you can run the code under defined/not define "CLOCK_MONOTONIC_COARSE" by command? I mean how you can pick the clock CLOCK_MONOTONIC_COARSE, as well as CLOCK_MONOTONIC to run by command?

I'm sorry, I don't quite understand what you're asking here. Could you give an example of how you want this to work? How you imagine that command could look?

danghai:
I pass `CLOCK_MONOTONIC` instead of `clock_id` at rate-limitting writer. Because `clock_nanosleep` does not support the `CLOCK_MONOTONIC_COARSE`. I also remove the `clock_id` through to that place.

spbnick:
Thank you, Hai, this looks good! We only need to work on the commit message a bit.

First, keep the message subject short, preferably under 50 characters, but a
bit longer is OK. Explain why the change is necessary and what it fixes in the
message body.

Put a "Fixes #XX" line at the end to close the corresponding issue
automatically when we merge the commit.
See https://github.com/blog/1386-closing-issues-via-commit-messages

spbnick:
Thank you, Hai! You got the idea of the commit subject and message right, but your English was difficult to understand, so I rewrote them. Merged now.

PR #181, opened 2018-02-14
Add clock_gettime return codes in tty_source

danghai:


lib/tlog/tty_source.c
```diff
@@ -160,7 +160,9 @@ tlog_tty_source_loc_get(const struct tlog_source *source)
         return 0;
     } else {
         struct timespec ts;
-        clock_gettime(tty_source->clock_id, &ts);
+        if (clock_gettime(tty_source->clock_id, &ts) < 0) {
+            return TLOG_GRC_ERRNO;
+        }
```
spbnick: OK, now I remember why there was no check here. Because the interface doesn't allow reporting failure. You need to look at the return type of this function - it's size_t, not tlog_grc, and then take a look at the interface at https://github.com/Scribery/tlog/blob/96ad318036587572e1811a0e5c60aa708bdf9245/include/tlog/source_type.h#L61

Ignoring error is "okayish" here, returning the correct location is not critical to operation, only for error reporting.

So, let's drop this PR.

danghai: Ah okay

PR #186, opened 2018-03-06
Update README with correct $programname for rsyslog

theblazehen:


spbnick:
Thank you, @theblazehen, merged!

PR #189, opened 2018-03-08
Make es_json_reader compatible with Elasticsearch 5

### LINKED ISSUE #188: Compatibility with Elasticsearch 5.x
`tlog-play` is incompatible with ES 5, as it uses the `fields` parameter in the query string. The `fields` parameter is always empty, so it may be safe to remove.

Error message:
```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"The parameter [fields] is no longer supported, please use [stored_fields] to retrieve stored fields or _source filtering if the field is not stored"}],"type":"illegal_argument_exception","reason":"The parameter [fields] is no longer supported, please use [stored_fields] to retrieve stored fields or _source filtering if the field is not stored"},"status":400}
```

theblazehen:
Tested, and working on Elasticsearch 2 and 5

Fixes #188

### DIFF
--- a/lib/tlog/es_json_reader.c
+++ b/lib/tlog/es_json_reader.c
@@ -128,7 +128,7 @@ tlog_es_json_reader_format_url_pfx(char **purl_pfx,
 #undef ESC
 
     if (asprintf(&url_pfx,
-                 "%s?q=%s&fields=&_source&sort=%s&size=%zu&from=",
+                 "%s?q=%s&sort=%s&size=%zu&from=",
                  base_url, esc_query, esc_sort, size) < 0) {
         grc = TLOG_GRC_FROM(errno, ENOMEM);
         goto cleanup;


lib/tlog/es_json_reader.c
```diff
@@ -128,7 +128,7 @@ tlog_es_json_reader_format_url_pfx(char **purl_pfx,
 #undef ESC
 
     if (asprintf(&url_pfx,
-                 "%s?q=%s&fields=&_source&sort=%s&size=%zu&from=",
```
spbnick: Hmm, was support for `_source` removed as well? Why do we need to remove it?

theblazehen: Ah, right. Was a bit overzealous with the removal there. Since we aren't specifying any value there, is there any reason to keep `_source` in the query string?

spbnick: I have no idea if it's even used by Elasticsearch. I can't find any documentation about it and don't remember why I added it there. I'll need to do a little testing on this. However, in any case, removing `_source` needs to be a separate commit (within this PR is OK), if you could do it.

spbnick: There seems to be no difference in Elasticsearch 2.4.6 response, at least, so I'll have to assume that these particular URL parameters came from some early prototyping code, where I didn't care much. Let's remove `_source` as well. Could you add a separate commit to this PR, something like "Remove unused _source parameter from ES query URL"?

spbnick:
Merged as part of #190. Thank you, @theblazehen!

Issue #182, opened 2018-02-23
Failed creating lock file 


danghai:
Every time I turn on the terminal, it shows the error: Failed creating lock file. 
![fail](https://user-images.githubusercontent.com/10912671/36576466-b3d9098a-1805-11e8-8035-ff91f425851a.JPG)

spbnick:
Read https://github.com/Scribery/tlog#installing regarding session lock directory.

danghai:
I already created `tlog` in /var/run and make it writable (only) for the user(s) tlog-rec-session runs as. But it still has error when I turn on terminal

spbnick:
This is because your session lock directory is configured to be `/usr/var/run/tlog`. Add `--localstatedir=/var` to configure options, and rebuild. Also, it would be good to add it to build instructions in README.md.

danghai:
It still has error when I try to use `--localstatedir=/var` to configure options.  The error is only removed when I use `sudo make uninstall`

spbnick:
You need to figure out which binary is actually executed when the user logged in and whether it has the updated path compiled in, or not.

PR #191, opened 2018-03-19
Add localstatedir=/var to configure option in README.md

danghai:
Pull request for issue #182 ,
As you mention, it is good to add `localstatedir=/var` to configure option. It resolve the lock file issue. 
fix_lock_file


spbnick:
Thank you, Hai, merged!

Issue #41, opened 2016-02-26
Make configuration validation function return error message


spbnick:
Make tlog-rec and the future tlog-play configuration validation function return validation error string instead of printing it, so the function could be used in assertions.

spbnick:
The first part of this was done with the introduction of error stacks, only the assertions need to be added now.

spbnick:
This means asserting that `tlog_*_conf_validate` functions return TLOG_RC_OK on entering functions accepting corresponding configurations, and on exiting functions modifying them.

danghai:
could you elaborate your idea? I see you check function `tlog_*_conf_validate` with return value `grc`. I am not sure about using `assertion` in here.

spbnick:
Functions extracting data from loaded configuration expect it to be correct. Adding an `assert` to the start of the function verifying that, will help us ensure that. Similarly, functions which produce, or modify configuration, which are supposed to return a valid configuration need such `assert` at the exit.

danghai:
I am little bit confused. the function `tlog_*_conf_validate` is used to check validation of configuration. It does not modify, or produce configuration. So I do not need to add `assert` in here. I put `assert` in the function that it is used. The commit shows an example about adding `assert` : 
https://github.com/danghai/tlog/commit/72205228a664b7d5006080f9c5d6e0bea2c62eac

spbnick:
Yeah, we don't need to assert things in `tlog_*_conf_validate` functions for this issue, but instead use those functions to validate the configuration in functions which create, access or modify it. The `tlog_play_conf_file_load` doesn't need an assert, because it's the job of that function to validate the configuration and produce an error message for the user if it's wrong.

However, e.g. `tlog_play_conf_load` might need an `assert(tlog_play_conf_validate(NULL, conf, TLOG_CONF_ORIGIN_ARGS) == TLOG_RC_OK)` at the successful exit. I would add `TLOG_CONF_ORIGIN_MAX` (or `_LAST`, or `_TOP`), and use that instead, though.

PR #193, opened 2018-03-22
Make configuration validation function return validation error message

danghai:
Hi Nick, PR for issue #41 to make configuration validation function return validation error message by using `assert`.
1. As you mentioned, you are going to add `TLOG_CONF_ORIGIN_MAX` (or `_LAST`, or `_TOP`), and use that instead, though. What do they represent for?
2. I am not sure about what difference between using `assert` and using function return `grc`?. In particular,
```c
assert(tlog_play_conf_validate(NULL, conf, TLOG_CONF_ORIGIN_ARGS) == TLOG_RC_OK);
``` 

and 

```c
    grc = tlog_play_conf_validate(perrs, conf, TLOG_CONF_ORIGIN_ARGS);
    if (grc != TLOG_RC_OK) {
        TLOG_ERRS_RAISEF("Invalid contents of \"%s\"", path);
    }
```

3. I am not sure about `lowest priority first`. What does it mean?
https://github.com/Scribery/tlog/blob/master/include/tlog/conf_origin.h
Thanks


danghai:
I have updated it

lib/tlog/play_conf.c
```diff
@@ -75,6 +75,9 @@ tlog_play_conf_load(struct tlog_errs **perrs,
     assert(pcmd_help != NULL);
     assert(pconf != NULL);
     assert(argv != NULL);
+    /* Check validate the config */
+    assert(tlog_play_conf_validate(NULL, conf, TLOG_CONF_ORIGIN_ARGS) ==
+            TLOG_RC_OK);
```
spbnick: This wouldn't make sense, we don't need to validate a NULL pointer we just initialized. This function doesn't need to validate the configuration it receives, as it doesn't receive any. The same for the rest of those.

PR #200, opened 2018-07-17
Added BuildDependencies to allow yum-builddep

### LINKED ISSUE #199: tlog.spec should contain all required dependencies for building on EL6 and EL7
Users should be able to use `yum-builddep` on a minimal system to install all necessary dependencies for building `tlog` from source.

trevor-vaughan:
On a minimal system, the existing dependencies do not provide enough
information to fully build tlog from source.

Adding a few items allows us to use `yum-builddep` and install all
required dependencies for both EL6 and EL7 systems.

Fixes #199

### DIFF
--- a/tlog.spec
+++ b/tlog.spec
@@ -8,13 +8,16 @@ License:    GPLv2+
 URL:        https://github.com/Scribery/%{name}
 Source:     https://github.com/Scribery/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz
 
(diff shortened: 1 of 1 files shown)


spbnick:
Thank you for the fix, Trevor! I cleaned it up a little and pushed as another PR (preserving your authorship). Sorry, I'm slow to pick these up, I was occupied with another project a lot for the past several months, but hopefully @justin-stephenson would soon be able to help.

PR #203, opened 2018-07-17
Add sample profile hooks for people to use

trevor-vaughan:
Users should be able to have tlog automatically log sessions when they
access a system.

These scripts provide a sample for users to work from and/or include in
automation scripts as they see fit.


trevor-vaughan:
Do not pull, this needs some fixes before it's ready due to Bash shenanigans.

trevor-vaughan:
This should be ready to go

PR #213, opened 2018-09-18
Modify comma placement in tlog configuration files

justin-stephenson:
This PR is intended to modify the comma character placement in the default generated configuration files.

To allow successful front-end JSON parsing of configuration files, remove the trailing comma from container blocks, such as:

```
   // Journal writer parameters
    "journal": {
        // The syslog-style priority "journal" writer should use for messages.
        // "priority" : "info",

        // If true, the "journal" writer copies the following JSON fields
        // to Journal fields: user -> TLOG_USER, session -> TLOG_SESSION,
        // rec -> TLOG_REC, and id -> TLOG_ID.
        // "augment" : true
    },

```

Add and utilize the M4_PARAM_LAST_FIELD macro in certain situations where no trailing comma is needed.

This modifies the tlog-rec and tlog-rec-session configuration files, but the PR is still a work in progress as it creates an empty container in the m4 generated tlog-play.conf files.
```

     // Systemd journal reader parameters                                                                                                                                                                        
     "journal": {                                                                                                                                                                                                
                                                                                                                                                                                                                 
     }        
```
I believe that when M4_CONTAINER(`', `/journal', `Systemd journal reader') is expanded the `M4_CONF_CONTAINER_SIZE` macro is returning != 0 and that is why the empty container gets created but I have not been able to determine the root cause.

### DIFF
--- a/m4/tlog/conf_misc.m4
+++ b/m4/tlog/conf_misc.m4
@@ -57,12 +57,14 @@ m4_define(
 `m4_pushdef(`M4_CONF_CONTAINER_SIZE_VAL', `0')m4_dnl
 m4_pushdef(`M4_CONTAINER', `')m4_dnl
 m4_pushdef(`M4_PARAM', m4_defn(`_M4_CONF_CONTAINER_SIZE_ADD_PARAM'))m4_dnl
+m4_pushdef(`M4_PARAM_LAST_FIELD', m4_defn(`_M4_CONF_CONTAINER_SIZE_ADD_PARAM'))m4_dnl
 m4_pushdef(`M4_CONF_PREFIX', `$2')m4_dnl
 m4_pushdef(`M4_CONF_ORIGIN', `$3')m4_dnl
 m4_include(`$1')m4_dnl
 m4_popdef(`M4_CONF_ORIGIN')m4_dnl
 m4_popdef(`M4_CONF_PREFIX')m4_dnl
 m4_popdef(`M4_PARAM')m4_dnl
+m4_popdef(`M4_PARAM_LAST_FIELD')m4_dnl
 m4_popdef(`M4_CONTAINER')m4_dnl
 M4_CONF_CONTAINER_SIZE_VAL()m4_dnl
 m4_popdef(`M4_CONF_CONTAINER_SIZE_VAL')'m4_dnl
--- a/m4/tlog/play_conf_schema.m4
+++ b/m4/tlog/play_conf_schema.m4
@@ -86,7 +86,7 @@ M4_PARAM(`/es', `baseurl', `file-',
          `M4_LINES(`base URL to request Elasticsearch through. Should not',
                    `contain query (?...) or fragment (#...) parts.')')m4_dnl
 m4_dnl
-M4_PARAM(`/es', `query', `file-',
+M4_PARAM_LAST_FIELD(`/es', `query', `file-',
          `M4_TYPE_STRING()', false,
          `', `=STRING', `Elasticsearch query',
          `STRING is the ', `The ',
--- a/m4/tlog/rec_common_conf_schema.m4
+++ b/m4/tlog/rec_common_conf_schema.m4
@@ -57,7 +57,7 @@ _M4_PARAM(`/log', `output', `file-',
           `If specified as ', `If ',
           `M4_LINES(`true, terminal output is logged.')')m4_dnl
 m4_dnl
-_M4_PARAM(`/log', `window', `file-',
+_M4_PARAM_LAST_FIELD(`/log', `window', `file-',
           `M4_TYPE_BOOL(true)', true,
           `', `[=BOOL]', `Enable/disable logging terminal window size changes',
           `If specified as ', `If ',
@@ -80,7 +80,7 @@ _M4_PARAM(`/limit', `burst', `file-',
           `M4_LINES(`number of bytes by which logged messages are allowed to exceed',
                     `the rate limit momentarily, i.e. "burstiness".')')m4_dnl
 m4_dnl
-_M4_PARAM(`/limit', `action', `file-',
+_M4_PARAM_LAST_FIELD(`/limit', `action', `file-',
           `M4_TYPE_CHOICE(`pass', `pass', `delay', `drop')', true,
           `', `=STRING', `Perform STRING action above limits (pass/delay/drop)',
           `STRING is the ', `The ',
@@ -130,7 +130,7 @@ _M4_PARAM(`/syslog', `facility', `file-',
           `STRING is the ', `The ',
           `M4_LINES(`syslog facility "syslog" writer should use for messages.')')m4_dnl
 m4_dnl
-_M4_PARAM(`/syslog', `priority', `file-',
+_M4_PARAM_LAST_FIELD(`/syslog', `priority', `file-',
           `M4_TYPE_CHOICE(`info',
                           `emerg',
                           `alert',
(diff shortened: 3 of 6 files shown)


justin-stephenson:
I fixed the issue with tlog-play, this should be working now.

justin-stephenson:
Thank you Nikolai, I look forward to it!

If it helps I will try to explain a bit about what I discovered in my analysis before submitting the PR, and please note that I may be making some incorrect assumptions so feel free to correct them.

Prior to the patches, expansions of M4_PARAM and M4_CONTAINER would add a comma character to the preceding section due to:

```
  m4_ifelse(
    M4_FIRST(),
    `true',
    `m4_define(`M4_FIRST', `false')',
    `m4_printl(`,', `')'
  )
```

This would cause a comma character to be added to the end of a previous block such as the comma at the end of the journal section here.
```

65     // Journal writer parameters                                                                                                                                                                                
66     "journal": {                                                                                                                                                                                                
67         // The syslog-style priority "journal" writer should use for messages.                                                                                                                                  
68         // "priority" : "info",                                                                                                                                                                                 
69                                                                                                                                                                                                                 
70         // If true, the "journal" writer copies the following JSON fields                                                                                                                                       
71         // to Journal fields: user -> TLOG_USER, session -> TLOG_SESSION,                                                                                                                                       
72         // rec -> TLOG_REC, and id -> TLOG_ID.                                                                                                                                                                  
73         // "augment" : true                                                                                                                                                                                     
74     },                                                                                                                                                                                                          
75                                                                                                                                                                                                                 
76     // The type of "log writer" to use for logging. The writer needs                                                                                                                                            
77     // to be configured using its dedicated parameters.                                                                                                                                                         
78     // "writer" : "file"   
```
 
This behavior seemed a bit confusing to me, so this patch modifies the comma character to be added conditionally to the end of the M4_PARAM and M4_CONTAINER calls. The added macro M4_PARAM_LAST_FIELD  can be used in some cases to exclude the comma. The addition of M4_PARAM_LAST_FIELD was necessary, for example to ensure the 'action: pass' line 47 below did not include a comma.
```

34     // Logging limit parameters                                                                                                                                                                                 
35     "limit": {                                                                                                                                                                                                  
36         // The maximum rate messages could be logged at, bytes/sec.                                                                                                                                             
37         // "rate" : 16384,                                                                                                                                                                                      
38                                                                                                                                                                                                                 
39         // The number of bytes by which logged messages are allowed to exceed                                                                                                                                   
40         // the rate limit momentarily, i.e. "burstiness".                                                                                                                                                       
41         // "burst" : 32768,                                                                                                                                                                                     
42                                                                                                                                                                                                                 
43         // The logging limit action.                                                                                                                                                                            
44         // If set to "pass" no logging limits will be applied.                                                                                                                                                  
45         // If set to "delay", logging will be throttled.                                                                                                                                                        
46         // If set to "drop", messages exceeding limits will be dropped.                                                                                                                                         
47         // "action" : "pass"                                                                                                                                                                                    
48     },      

```

spbnick:
Thank you for the explanation, Justin, I was missing it :) It would be good to know what exact issue you were trying to solve as well. I'm assuming you're trying to fix the case where there would be a trailing comma in a JSON container, in a configuration file, which is considered invalid by the official specification and some readers. I also assume that the requirement comes from @sabbaka.

I have to note, from the start, that your code produces invalid JSON. E.g. see the missing comma in `tlog-play.default.conf`. However, there are more things to consider.

The current configuration is generated with the expectation that the reader would be able to handle trailing commas. I also think that the requirement of no trailing commas is a mistake, in the original official specification. Yes, they wanted to make it simple, but they shouldn't have. Permitting trailing commas in containers is the widespread complexity tradeoff between parsers and generators. Many readers ignore that, or provide an option to ignore that. @sabbaka, could you please explain what requires the absence of the trailing commas, and could you please check if there is a way to make it accept them, or perhaps use a different reader?

The existing code actually handles removing trailing commas, and that's what the `M4_FIRST` macro is for. It is set to `true` at the start of every container (including top-level), and is set to `false` as soon as the first item in the container is output. It is also used to decide whether to place the comma at the end of the previous item: if it is `true`, it means there was no item in the container yet, and the comma is not added, and if it's `false`, then the comma is added.

If we added the comma *after* the item, we would need to know if each item is the last one in the container (which you had to do with your `M4_PARAM_LAST_FIELD` macro), and that makes the code and the lists more difficult to maintain, and more prone to mistakes when editing. Moreover it doesn't solve the problem which brought this PR in the first place.

The problem is that some items need to be commented-out in the configuration installed into `/etc` to show that the defaults are used instead, and to show what those defaults are. Such approach is customary for configuration files. As a result, if the last (but not the only) item in a container is commented-out, the previous item would have a trailing comma attached to it. The same would happen with either approaches: adding comma in front, or at the back of an item.

This, however, works OK with most of JSON readers, as they handle trailing commas.

The alternative could be to not add trailing comma to items which are not followed by uncommented items, like this:

```
[
    10,
    20
]
```
or this:
```
[
    10,
    20
    // 30
    // 40
]
```
 That is, of course, possible to implement. However, then, when users uncomment the trailing items to change the setting, JSON becomes invalid, breaking the configuration:
```
[
    10,
    20
    30
    // 40
]
```
Of course, we can say that users should watch out for that, but such mistake is easy to make.

So, I chose to rely on a flexible JSON reader instead of setting a trap for the users. Especially since we already use comments anyway, which are also non-standard and shouldn't be accepted by JSON readers.

sabbaka:
By the JSON specification trailing comas are not allowed. So, I believe it is fair enough to fix the part which is broken. In this case this is configuration file of tlog. I am using library to work with configuration file of tlog. Library name is comment-json. In addition to that, comments are also not allowed by JSON specification. Basically, what we are doing here from both sides is sort of crutch. In such, I would say, complex project it is a sacrifice in some way. Currently in our schedule I would prefer not to search ( or modify existing ) for another JavaScript library which will not only support JSON comments, but also trailing comas. It seems it would easier to just remove that one trailing coma.
Also, config file which will be generated by comment-json will not include comments at all. So, maybe it is better to have all parameters uncommented from the beginning. 
I would like to find a better solution here...
Let's discuss this and find the best solution. From my side I can look for something else or maybe try to keep comments and somehow keep the line breaks and spaces.

spbnick:
@sabbaka, I see that you would rather not change the library you already use at this moment. I think I remember that we also had a problem of merging the defaults when reading tlog configuration files. The solution was to rely on tlog itself to output the *actual* configuration it is using, already merged, and let the Cockpit UI read *that* instead.

How about we implement an option to `tlog-rec`, `tlog-rec-session`, and `tlog-play` which makes them output the actual configuration they use (accounting for the defaults), and do it in strictly-conformant JSON, without comments and trailing commas? Then instead of reading the files, you would be able to execute the commands, read their output, and be sure you read the configuration correctly? You would be able to write the exact JSON you need into /etc/tlog and tlog will handle it.

sabbaka:
That sounds great. I think that is a great idea. What do you think about overwriting the configuration file upon saving? Could we make then a way to save configuration as well? 

As I see it - tlog-rec-session --configuration is called and returns a json object with all current settings. Then tlog-rec-session --set-configuration is called and as an input json is provided. Are you thinking about something like this?

spbnick:
No we would only have tlog-rec-session print the configuration. Then you would take the original values of configuration settings from there, modify them and write only those you're interested in in JSON to /etc/tlog/tlog-rec-session.conf. You can also add a nice header comment explaining e.g. where the settings come from. Next time you ask tlog-rec-session to dump the configuration it would read that.

sabbaka:
That is a great idea which I definitely support. Also I will try to find a way to preserve original comments, though it might be tricky. We'll see.

Thank you.

spbnick:
There would be non need to preserve original comments, as there would be no comments in the configuration output with "--configuration" (and the JSON library tlog uses doesn't support preserving them anyway).

I created #217 to implement "--configuration" option, and fix the command option parsing (another issue).

spbnick:
Closing this PR. Please reopen if you feel we should still merge this in one shape or another. Thank you, @justin-stephenson and @sabbaka!

sabbaka:
@justin-stephenson thank you for the effort, I hope you won't be disappointed that it was not merged in, but I believe we have found a great solution here.  Thanks to @spbnick

justin-stephenson:
No problem, thank you for the information.

PR #217, opened 2018-09-21
Command option parsing and --configuration option

spbnick:
This switches command-line option parsing to stop looking for options after the first non-option argument, and adds the "--configuration" option.

This supersedes #213.

### DIFF
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -1187,6 +1187,23 @@ tlog_play(struct tlog_errs **perrs,
         }
     }
 
+    /* Check for the configuration flag */
+    if (json_object_object_get_ex(conf, "configuration", &obj)) {
+        if (json_object_get_boolean(obj)) {
+            const char *str;
+            str = json_object_to_json_string_ext(conf,
+                                                 JSON_C_TO_STRING_PRETTY);
+            if (str == NULL) {
+                grc = TLOG_GRC_ERRNO;
+                TLOG_ERRS_RAISECS(grc,
+                                  "Failed formatting configuration JSON");
+            }
+            printf("%s\n", str);
+            grc = TLOG_RC_OK;
+            goto cleanup;
+        }
+    }
+
     /* Output and discard any accumulated non-critical error messages */
     tlog_errs_print(stderr, *perrs);
     tlog_errs_destroy(perrs);
--- a/lib/tlog/rec.c
+++ b/lib/tlog/rec.c
@@ -1107,6 +1107,23 @@ tlog_rec(struct tlog_errs **perrs, uid_t euid, gid_t egid,
         }
     }
 
+    /* Check for the configuration flag */
+    if (json_object_object_get_ex(conf, "configuration", &obj)) {
+        if (json_object_get_boolean(obj)) {
+            const char *str;
+            str = json_object_to_json_string_ext(conf,
+                                                 JSON_C_TO_STRING_PRETTY);
+            if (str == NULL) {
+                grc = TLOG_GRC_ERRNO;
+                TLOG_ERRS_RAISECS(grc,
+                                  "Failed formatting configuration JSON");
+            }
+            printf("%s\n", str);
+            grc = TLOG_RC_OK;
+            goto cleanup;
+        }
+    }
+
     /*
      * Choose the clock: try to use coarse monotonic clock (which is faster),
      * if it provides the required resolution.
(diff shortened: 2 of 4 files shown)


spbnick:
@justin-stephenson, would you mind doing a quick review of this?

sabbaka:
That seems to be okay for me. Anyway, I will be creating some sort of serializer which handle data in and data out properly.

justin-stephenson:
Thank you for the quick fixes @spbnick 

I did a test and noticed two things, one is that the following error occurs when redirecting tlog-rec --configuration output:

```
# tlog-rec --configuration > /etc/tlog/tlog-rec.conf
unexpected end of data
Failed loading "/etc/tlog/tlog-rec.conf"
Failed loading system configuration
# file /etc/tlog/tlog-rec.conf
/etc/tlog/tlog-rec.conf: empty
```

This however works without issues:

`# tlog-rec --configuration > ~/tlog-rec.conf`

Also, the printed --configuration output does not remove the trailing container comma which is(my understanding) the reason that @sabbaka raised the original problem, i'd like to have him comment on this however. For example:


```
# tlog-rec --configuration
 ...
<snip>
...
  "journal":{
    "priority":"info",
    "augment":true
  },

```

sabbaka:
"journal":{
    "priority":"info",
    "augment":true
  },

Is that all output? After "augment":true there is no coma which is correct.

justin-stephenson:
Apologies, perhaps I misunderstood the original issue. Here is the full output:

```
# tlog-rec --configuration
{
  "latency":10,
  "payload":2048,
  "writer":"file",
  "log":{
    "input":false,
    "output":true,
    "window":true
  },
  "limit":{
    "rate":16384,
    "burst":32768,
    "action":"pass"
  },
  "file":{
  },
  "syslog":{
    "facility":"authpriv",
    "priority":"info"
  },
  "journal":{
    "priority":"info",
    "augment":true
  },
  "configuration":true,
  "args":[
  ]

```

spbnick:
@justin-stephenson, this command:

    tlog-rec --configuration > /etc/tlog/tlog-rec.conf

fails, because when shell sets up the redirection to `/etc/tlog/tlog-rec.conf` it opens it with O_TRUNC, which empties the file. That happens before `tlog-rec` is even started, and has nothing to do with it, so when it finally starts it sees an empty file. That's just the way shell works and is not a bug.

spbnick:
Also, @sabbaka's problem was trailing commas in containers in general, not just a comma after that specific one. I.e. for his reader this wouldn't work: `{"a":1, "b":2, "c":3,}`, but this would: `{"a":1, "b":2, "c":3}`

spbnick:
Are there any other issues you see, @justin-stephenson?

justin-stephenson:
Looks good to me, thank you @spbnick

spbnick:
Thanks, Justin! Merging.

sabbaka:
The `tlog-rec-session --configuration` works, but there is an error which appears after.

```
[kgliebov@localhost tlog]$ tlog-rec-session --configuration
{
  "shell":"\/bin\/bash",
  "notice":"\nATTENTION! Your session is being recorded!\n\n",
  "latency":10,
  "payload":2048,
  "log":{
    "input":false,
    "output":true,
    "window":true
  },
  "limit":{
    "rate":16384,
    "burst":32768,
    "action":"pass"
  },
  "file":{
  },
  "syslog":{
    "facility":"authpriv",
    "priority":"info"
  },
  "journal":{
    "priority":"info",
    "augment":true
  },
  "writer":"journal",
  "configuration":true,
  "args":[
  ]
}
[kgliebov@localhost tlog]$ sudo vim /etc/tlog/tlog-rec-session.conf
[kgliebov@localhost tlog]$ tlog-rec-session --configuration
Unexpected node: "configuration"
Failed validating contents of "/etc/tlog/tlog-rec-session.conf"
Failed loading system configuration
```

Same goes for configuration saved from Cockpit:

```
[kgliebov@localhost tlog]$ cat /etc/tlog/tlog-rec-session.conf
{"shell":"/bin/bash","notice":"\nATTENTION! Your session is being recorded!\n\n","latency":10,"payload":2048,"log":{"input":false,"output":true,"window":true},"limit":{"rate":16384,"burst":32768,"action":"pass"},"file":{},"syslog":{"facility":"authpriv","priority":"info"},"journal":{"priority":"info","augment":true},"writer":"journal","configuration":true,"args":[]}
[kgliebov@localhost tlog]$ tlog-rec-session --configuration
Unexpected node: "configuration"
Failed validating contents of "/etc/tlog/tlog-rec-session.conf"
Failed loading system configuration

```

spbnick:
@sabbaka, that's what I described [above](https://github.com/Scribery/tlog/pull/217#issuecomment-423486712).

PR #229, opened 2019-03-12
CI: Fix coverage analysis

justin-stephenson:
This PR should fix some issues with coverage analysis evident on https://coveralls.io/github/Scribery/tlog

Execute `src/tlog/tlog-{play,rec,rec-session}` programs in Travis CI, this is required
to generate gcda files to include **src/tlog/*.c** files in coverage reports.

Exclude unneeded subdirectories from coverage analysis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,4 +34,4 @@ script:
 - rpmbuild --define "_tmpfilesdir /usr/lib/tmpfiles.d" -ta --nodeps tlog-*.tar.gz
 - test "x$DISTCHECK" = xyes && make distcheck || true
 after_success:
-- test "x$COVERAGE" = xyes -a "x$CC" = "xgcc" && lcov --compat-libtool --directory . --capture --output-file coverage.info && coveralls-lcov coverage.info
+- test "x$COVERAGE" = xyes -a "x$CC" = "xgcc" && lcov --compat-libtool --directory ./lib/tlog/ --capture --output-file coverage.info && coveralls-lcov coverage.info


coverage.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+echo "Executing coverage analysis script"
+# Programs below must be executed for gcov to generate .gcda files
+./src/tlog/tlog-rec-session -c whoami
+./src/tlog/tlog-rec
+./src/tlog/tlog-play
```
spbnick: Coverage data by itself is worthless. Coverage data is a (rough) metric of how much of your program's functionality is being tested. Just starting the programs and not checking what they do, not even checking what their exit status is, would produce some coverage data (i.e. we executed some code) but it wouldn't mean that we tested anything, because we haven't actually checked the outcome.

So, please remove these three lines, write some actual tests, e.g. with Scott's help, and make them run in Travis CI.

coverage.sh
```diff
@@ -0,0 +1,9 @@
...
+lcov --compat-libtool --directory . --capture --output-file coverage.info
+lcov --remove coverage.info '*src/tltest*' '*lib/tltest*' '*include/tlog*' '*include/tltest*' '*usr/include/curl*' -o coverage.info
+coveralls-lcov coverage.info
```
spbnick: AFAIK, nothing except Travis CI would find this code useful and there's not much of it, so please don't add another file in the top directory and keep these lines inside .travis.yml.

coverage.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+echo "Executing coverage analysis script"
+# Programs below must be executed for gcov to generate .gcda files
+./src/tlog/tlog-rec-session -c whoami
+./src/tlog/tlog-rec
+./src/tlog/tlog-play
+lcov --compat-libtool --directory . --capture --output-file coverage.info
+lcov --remove coverage.info '*src/tltest*' '*lib/tltest*' '*include/tlog*' '*include/tltest*' '*usr/include/curl*' -o coverage.info
```
spbnick: Instead of collecting coverage data from all directories, and then removing data for some directories, could you instead collect it only from the directories we need? I.e. write 'lcov --compat-libtool --directory lib/tlog --directory src/tlog ...`?

justin-stephenson:
Thanks for the review @spbnick.

I updated the PR to only exclude the unnecessary directories and addressed the other changes you commented about.

As we are now only getting started adding integration tests, I propose we can add tlog-* binary coverage analysis at a later point in time once there is a need to analyze integration test coverage. If you agree I will create an issue to track this and it will be excluded from this PR.

spbnick:
Agreed :) The sooner we get some integration testing going in Travis CI, the better :)

Issue #222, opened 2018-10-30
Consider reporting Journal not present


spbnick:
When Journal is not installed on the system, sd_journal_send would return ENOENT as seen in #221. Consider detecting that and printing an appropriate error message.

justin-stephenson:
This issue can be closed now, fixed in the referenced PR.

PR #230, opened 2019-03-12
Log more detailed error when systemd journal is not present

justin-stephenson:
On distributions where `systemd` is not installed(or `systemd-journal` is not available),
logging to the journal will inevitably fail. Improve the error message logged when
this occurs. This PR is created for Issue #222 

The resulting error message looks like:
```
No such file or directory
Systemd journal not available or responsive
Failed transferring TTY data
```

I expect there is a better way to handle this, please review :)


justin-stephenson:
Fixed the wording, thanks !

spbnick:
Thank you :)

PR #231, opened 2019-03-13
Add initial tlog integration test

spoore1:
This commit adds the first integration tests.  These tests are written
in python for pytest execution.  More specific information about how to
setup for the tests, run them, and cleanup afterwards is described in
the README.md file.

This commit provides:

* README.md -- descibes the tests and how to use them
* tlitest-setup -- script to setup dependencies like rpms and users
* tlitest-teardown -- script to cleanup after running the tests
* test_tlog.py -- the main pytest module for tlog integration tests

### DIFF
--- a/src/tlitest/README.md
+++ b/src/tlitest/README.md
@@ -1,0 +1,51 @@
+Tlog Integration Tests
+======================
+
+Tlog Integration Tests provide real world tests for various features
+and functionalities of tlog.  Although these tests are meant for
+OS package build gating, they can be run for general testing purposes
+as well.
+
+This README.md will provide information on how to prepare for, run,
+and cleanup after the tests.
+
+Setup
+-----
+
+In the src/tlitest directory, run the main setup script:
+    ./tlitest-setup
+
+This will install necessary packages, add users and setup a wheel
+based no password sudoer rule file.
+
+As a part of setup, we also take note of the specified packages that
+were already installed before setup.  This allows us to skip them
+during package removal of cleanup (teardown).
+
+Also, during setup, we take note of users the same way so that
+pre-existing users will skip delete during cleanup.
+
+Run
+---
+
+To run the tests, you must execute pytest.  In the src/tlitest
+directory run:
+    ./tlitest-run
+
+If you want to generate a junit xml file for upload for reporting,
+run:
+    ./tlitest-run --junit-xml=/tmp/path-to-your-junit-xml-file
+
+Cleanup
+-------
+
+The cleanup after the tests, inside the src/tlitest directory run:
+    ./tlitest-teardown
+
+This will look for pre-existing packages and skip them.  Packages
+installed by setup will be removed.
+
+This will also look for pre-existing users and skip them. Users
+added by setup will be removed.
+
+The sudoers rule file added by setup will also be removed.
--- a/src/tlitest/tlitest-run
+++ b/src/tlitest/tlitest-run
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+pytest-3 -v "$(dirname $0)/../../lib/tlitest" "$@"
--- a/lib/tlitest/test_tlog.py
+++ b/lib/tlitest/test_tlog.py
@@ -1,0 +1,129 @@
+""" tlog tests """
+
+import os
+import ast
+import stat
+import time
+import pytest
+import pexpect
+import inspect
+from pexpect import pxssh
+from systemd import journal
+from tempfile import mkdtemp
+
+
+def journal_find_last():
+    j = journal.Reader()
+    j.seek_tail()
+    entry = j.get_previous()
+    while('TLOG_REC' not in entry):
+        entry = j.get_previous()
+    return entry
+
+
+def check_journal(pattern):
+    time.sleep(1)
+    entry = journal_find_last()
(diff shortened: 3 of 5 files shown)


src/tlitest/README.md
```diff
@@ -0,0 +1,51 @@
...
+as well.
+
+This README.md will provide information on how to prepare for, run,
+and cleanup after the tests.
+
+Setup
+-----
+
+In the src/tlitest directory, run the main setup script:
+    sh ./tlitest-setup
```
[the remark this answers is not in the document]
spoore1: removed in latest update

src/tlitest/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+If you want to generate a junit xml file for upload for reporting,
+run:
+    pytest-3 -v --junit-xml=/tmp/path-to-your-junit-xml-file
+
+Cleanup
+-------
+
+The cleanup after the tests, inside the src/tlitest directory run:
+    sh ./tlitest-teardown
```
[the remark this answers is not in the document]
spoore1: done as well

src/tlitest/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Also, during setup, we take note of users the same way so that 
+pre-existing users will skip delete during cleanup.
+
+Run
+---
+
+To run the tests, you must execute pytest.  In the src/tlitest
+directory run:
+    pytest-3 -v
```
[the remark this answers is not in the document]
spoore1: I added a tlitest-run script and moved test_tlog.py to lib/tlitest and changed the README.md to reflect the change to use tlitest-run executable

src/tlitest/tlitest-run
```diff
@@ -0,0 +1,26 @@
...
+echo "RUN_PATH=$PWD"
+
+TEST_DIRS="
+../../lib
+/usr/local/lib
+/lib
+"
+
+for TEST_DIR in $TEST_DIRS; do
+    TEST_FILE=$(find $TEST_DIR -name test_tlog.py|head -1)
```
spbnick: You don't need to loop over the directories yourself, just pass all of them to `find`.
I.e.:

    find $TEST_DIRS -name test_tlog.py

src/tlitest/tlitest-run
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+cd $(dirname $0)
+echo "RUN_PATH=$PWD"
+
+TEST_DIRS="
+../../lib
+/usr/local/lib
+/lib
+"
```
spbnick: We don't really know where we'll be installed. To be handled reliably configure would need to substitute its install location here. However, we don't need to handle this now, if ever. We already ensured our executable names are globally-unique, and that's enough for now.

src/tlitest/tlitest-run
```diff
@@ -0,0 +1,26 @@
...
+    fi
+done
+
+if [ -z "$TEST_FILE" ]; then
+    echo "Cannot find test_tlog.py pytest file"
+    exit 1
+fi
+
+echo "Found $TEST_FILE...changing directory to run pytest"
+cd $(dirname $TEST_FILE)
```
spbnick: There doesn't seem to be a need to change to the directory to run a test with pytest, you can just supply it with a file/dir path.

src/tlitest/tlitest-run
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
spbnick: Do not constrain yourself to `/bin/sh` unnecessarily, as it's not always Bash there (especially on Debian/Ubuntu). Just write `/bin/bash` always, unless you really intend to limit yourself to POSIX shell features only.

src/tlitest/tlitest-run
```diff
@@ -0,0 +1,26 @@
...
+done
+
+if [ -z "$TEST_FILE" ]; then
+    echo "Cannot find test_tlog.py pytest file"
+    exit 1
+fi
+
+echo "Found $TEST_FILE...changing directory to run pytest"
+cd $(dirname $TEST_FILE)
+pytest-3 -v test_tlog.py $*
```
spbnick: Another thing: using `$*` would break as soon as you have whitespace in your arguments, which is a nasty surprise. So, in general use "$@", which will reproduce the arguments faithfully.

src/tlitest/tlitest-run
spoore1: That's very useful to know.

src/tlitest/tlitest-run
spoore1: Ok.

src/tlitest/tlitest-run
spoore1: You're right, executing with path should work.  And since this will initially be used from the source tree, I'll switch the entire file to the suggested one liner pytest call.

src/tlitest/tlitest-run
spoore1: Good point about other distros.  I wasn't thinking about that.  I'l switch all the scripts to specify bash.

src/tlitest/tlitest-run
spoore1: Forgot about $@.  I will switch as suggested.  Thanks

PR #234, opened 2019-03-15
Add pycache and py[co] to gitignore

spoore1:
Since the tlitest code uses pytest, we can potentially have additional
python files generated in source such as __pycache__/, *.pyc, or *.pyo
fiels.  We can ignore those from git.

Also, this was requested during review of PR #231.


spbnick:
Thank you, Scott!

PR #235, opened 2019-03-18
Fix signedness comparison errors

justin-stephenson:
Address the type comparison errors when `--enable-debug` configure option is set:

```
play_conf_validate.c: In function ‘tlog_play_conf_validate_journal’:
play_conf_validate.c:150:27: error: comparison of integer expressions of different signedness:
 ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Werror=sign-compare]
             for (i = 0; i < json_object_array_length(obj); i++) {
```


spbnick:
Ah, so json-c finally switched the return type of `json_object_array_length()` to `size_t`, good. So, could you instead switch the type of `i` to `size_t`, and cast the return value of `json_object_array_length()` to it (for older json-c versions)? In practice that doesn't make much difference, but it would be safer in theory, and clearer when reading.

justin-stephenson:
Updated, thanks for the review!

lib/tlog/conf_validate.c.m4
```diff
@@ -160,12 +160,12 @@ m4_define(
...
            `            }',
-           `            int i;',
-           `            for (i = 0; i < json_object_array_length(obj); i++) {',
+           `            size_t i;',
+           `            for (i = 0; i < (size_t)json_object_array_length(obj); i++) {',
            `                struct json_object *item_obj = json_object_array_get_idx(obj, i);',
            `                enum json_type item_type = json_object_get_type(item_obj);',
            `                if (item_type != json_type_string) {',
-           `                    tlog_errs_pushf(perrs, "Invalid \"%s\" item #%d type: %s",',
+           `                    tlog_errs_pushf(perrs, "Invalid \"%s\" item #%ld type: %s",',
```
spbnick: The format specifier for size_t should be `%zu` to match size_t better. See printf(3).

justin-stephenson:
Fixed the format specifier.

PR #236, opened 2019-03-18
Add more recording tests

spoore1:
Adding more tests for:

    test_record_interactive_session
    test_record_binary_output
    test_record_diff_char_sets
    test_record_fast_input
    test_record_as_unprivileged_user
    test_record_as_admin_user
    test_record_from_different_hostnames

Also adding necessary setup changes to add admin user to wheel group

### DIFF
--- a/lib/tlitest/test_tlog.py
+++ b/lib/tlitest/test_tlog.py
@@ -28,25 +28,29 @@ def journal_find_last():
 def check_journal(pattern):
     """ Check that last journal entry contains pattern """
     time.sleep(1)
-    entry = journal_find_last()
-    message = entry['MESSAGE']
-    out_txt = ast.literal_eval(message)['out_txt']
-    retval = 0
-    if pattern not in out_txt:
-        retval = 1
-    return retval
+    for _ in range(0, 10):
+        entry = journal_find_last()
+        message = entry['MESSAGE']
+        out_txt = ast.literal_eval(message)['out_txt']
+        if pattern in out_txt:
+            break
+        else:
+            time.sleep(5)
+    assert pattern in out_txt
 
 
 def check_outfile(pattern, filename):
     """ Check that file contains pattern """
     time.sleep(1)
-    file1 = open(filename, 'r')
-    content = file1.read()
-    file1.close()
-    retval = 0
-    if pattern not in content:
-        retval = 1
-    return retval
+    for _ in range(0, 10):
+        file1 = open(filename, 'r')
+        content = file1.read()
+        file1.close()
+        if pattern in content:
+            break
+        else:
+            time.sleep(5)
+    assert pattern in content
 
 
 def check_recording(shell, pattern, filename=None):
--- a/lib/tlitest/test_tlog.py
+++ b/lib/tlitest/test_tlog.py
@@ -20,9 +20,11 @@ def journal_find_last():
     j = journal.Reader()
     j.seek_tail()
     entry = j.get_previous()
-    while 'TLOG_REC' not in entry:
+    while 'TLOG_REC' not in entry and len(entry) != 0:
         entry = j.get_previous()
-    return entry
+    if 'TLOG_REC' in entry:
+        return entry
+    raise ValueError('Did not find TLOG_REC entry in journal')
 
 
 def check_journal(pattern):
--- a/lib/tlitest/test_tlog.py
+++ b/lib/tlitest/test_tlog.py
@@ -4,9 +4,11 @@ import os
 import ast
 import stat
 import time
+import socket
 import pytest
 import pexpect
 import inspect
+from shutil import copyfile
 from pexpect import pxssh
 from systemd import journal
 from tempfile import mkdtemp
@@ -87,6 +89,16 @@ def mklogfile(directory, filename=None):
     return logfile
 
 
+def mkcfgfile(filename, content):
+    """ helper to create config file from string """
+    bkup = '{}.origtest'.format(filename)
+    content = '\n'.join([line.lstrip() for line in content.split('\n')])
+    if os.path.isfile(filename) and not os.path.isfile(bkup):
+        copyfile(filename, bkup)
+    with open(filename, 'w') as out:
+        out.write(content)
+
+
 class TestTlogRec(object):
     tempdir = mkdtemp(prefix='/tmp/TestTlogRec.')
     user1 = 'tlitestlocaluser1'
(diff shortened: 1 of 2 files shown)


spoore1:
Found glibc-all-langpacks was missing from fedora build I tested on so adding it to tlitest-setup here as well.  rerunning local tests now to post to help with review.

spoore1:
Also found fedora didn't like the socket.sethostname by itself for journal hostname change test.  Added a write to /etc/hostname to fix the issue.

spoore1:
[tlitest-rectest-fedora29.log](https://github.com/Scribery/tlog/files/2993094/tlitest-rectest-fedora29.log)

Attaching output from a manual run on a Fedora 29 vm.

spoore1:
@spbnick @justin-stephenson can one of you guys review this?

lib/tlitest/test_tlog.py
```diff
@@ -2,48 +2,56 @@
...
     message = entry['MESSAGE']
     out_txt = ast.literal_eval(message)['out_txt']
-    if pattern in out_txt:
-        return 0
-    else:
-        return 1
+    retval = 0
+    if pattern not in out_txt:
+        retval = 1
+    return retval
```
spbnick: You can condense the above four lines into one:

    return pattern not in out_txt

Regardless of how you do it, I see you're returning the result of the comparison, but nothing checks for it, so no test would fail, no matter which result you get. Is that your intention?

The above concerns `check_outfile` as well.

lib/tlitest/test_tlog.py
```diff
@@ -2,48 +2,56 @@
...
 
 def journal_find_last():
+    """ helper to get the last journal entry """
     j = journal.Reader()
     j.seek_tail()
     entry = j.get_previous()
-    while('TLOG_REC' not in entry):
+    while 'TLOG_REC' not in entry:
         entry = j.get_previous()
     return entry
```
spbnick: This will return an entry, even if there are no tlog-rec entries in the journal and that might get confusing. I would return `None` if an entry is not found.

spoore1:
@spbnick 

Your suggestions all make sense.  Just to clarify, do you mean split them up into separate logical commits in this PR or separate PRs?

spbnick:
@spoore1, I mean split into separate commits **inside** this PR.

justin-stephenson:
I agree with @spbnick 's comments, in additional from my perspective it would be nice to have tests located in different files based on what is being tested. We may end up with a large quantity of integration tests therefore it would be nice to keep the tests logically seperated.

I wonder if we should have them organized by writer?
```

test_record_file.py 
test_record_journal.py
test_record_syslog.py
```

or something more categorical:
```

test_record_binary.py containing
  - test_record_binary_output

test_record_io_fuzzing.py
test_record_users.py - containing
  - test_record_as_unprivileged_user
  - test_record_as_admin_user
test_record_logging.py - containing
  - test_record_to_journal
  - test_record_to_syslog
test_record_misc.py - for others that may not be obvious
```

I had in mind something similar to organization like [SSSD tests](https://github.com/SSSD/sssd/tree/master/src/tests/intg)

Also what about keeping shared test utility functions like check_recording and check_outfile in a different .py file that can be imported?

spoore1:
@justin-stephenson 

I was thinking of separating them by command first and additional functional tests second.  The original tests I started with are organized into separate classes that I could easily beaker into separate files.

This is what I have now that I'm working to submit here:

class TestTlogRec:
class TestTlogRecPerformanceOptions:
class TestTlogPlay:
class TestTlogPlayRemote:
class TestTlogRecSession:

So, would it make sense to break up into files like that?

justin-stephenson:
@spoore1 that sounds good to me!

lib/tlitest/test_tlog.py
spoore1: Latest commit added changes to how check_journal and check_output work.   IIRC my original intention was to assert but that wasn't being done in the tests.  Looking at how it's used, I've instead moved the assert I originally intended into each function.  I also found failures in tests just today that I didn't see before so I changed the functions to loop a few times and check again.

lib/tlitest/test_tlog.py
spoore1: Actually in my testing a little while ago it looped indefinitely which is probably worse.

I added another commit to modify to code to add a conditional for len(entry) != 0 to while and raise an exception if TLOG_REC not found.  There may be cases later where we might use this to search for no tlog-rec entries in a journal but, right now this is only used for cases where they exist.  So, if they don't exist, probably best to raise an exception to fail the test case?

spoore1:
@spbnick 

I think I covered your requests.  I split up the commits and I think made the changes you suggested (or similar functionally at least I hope).

@justin-stephenson 

Can you take a look at the commits now?  I also included a final one marking the first three tests as tier1 that we can use in gating later like we discussed earlier.

justin-stephenson:
@spoore1 looks okay but I did not see the tests separated out into different .py files, can you double-check? I was thinking something like:
~~~
lib/tlitest/test_rec.py
lib/tlitest/test_rec_options.py
lib/tlitest/test_play.py
lib/tlitest/test_play_remote.py
lib/tlitest/util.py
~~~

spoore1:
@justin-stephenson 
I haven't added the other tests yet.  Thought I'd do that as a separate PR after this one was complete.

justin-stephenson:
@spoore1 okay got it, thanks for your contributions.

justin-stephenson:
@spoore1 I am seeing some intermittent failures running the tests on my system, 'test_record_diff_char_sets' is a test that is consistently failing.

~~~
self = <test_tlog.TestTlogRec object at 0x7fe903b9fdd8>

    def test_record_diff_char_sets(self):
        """
            Check tlog-rec preserves non-English I/O in recordings
            """
        logfile = '{}-ru_RU'.format(mklogfile(self.tempdir))
        shell = ssh_pexpect(self.user1, 'Secret123', 'localhost')
        shell.sendline('tlog-rec -o {} /bin/bash'.format(logfile))
        shell.sendline('export LANG=ru_RU.utf8')
        shell.sendline('badcommand')
        shell.sendline('exit')
>       check_outfile('найдена', logfile)

lib/tlitest/test_tlog.py:216:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

pattern = 'найдена', filename = '/tmp/TestTlogRec.ad_b1gvi/test_record_diff_char_sets.tlog-ru_RU'

    def check_outfile(pattern, filename):
        """ Check that file contains pattern """
        time.sleep(1)
        for _ in range(0, 10):
            file1 = open(filename, 'r')
            content = file1.read()
            file1.close()
            if pattern in content:
                break
            else:
                time.sleep(5)
>       assert pattern in content
E       assert 'найдена' in '{"ver":"2.2","host":"agalloch","rec":"93b093fa929e4c039c0be2a24d6b9421-3280-88cbe","user":"tlitestlocaluser1","term":...\n\\u001bktlitestlocaluser1@agalloch:~\\u001b\\\\[tlitestlocal
user1@agalloch ~]$ exit\\r\\nexit\\r\\n","out_bin":[]}\n'

lib/tlitest/test_tlog.py:55: AssertionError
~~~

Also, do you think it would be more efficient to use the subprocess module for recording tests and calling tlog-rec directly, instead of using pexpect? That way we could remove the overhead of SSH'ing to the system for each test, this might also address any test failures that may be timing related, or due to a race condition. I think you have more experience than I with python testing however, so i'll defer to your judgement.

 If I remember correctly you could only use pexpect for tlog-play tests, but perhaps we could isolate those.

spoore1:
I'm assuming you ran tlitest-setup from this change?  So you have glibc-all-langpacks installed?

I don't see out_txt in the AssertionError output.  Are you able to see the string in the file?

What other intermittent failures are you seeing?  I thought I'd fixed all the ones I found. What OS are you testing on?  Fedora 29?

As for subprocess vs pexpect, I did look into subprocess briefly when I started this a few months back.  
I chose pexpect mostly because it seemed to offer a simple interface to handle the interactive testing.  
It is my understanding that to use subprocess, we'd have to basically write our own expect() and sendline() functions to handle processing the I/O.

That being said, I'm not against looking into subprocess because it may indeed make testing in containers easier.  Can we write that up as an issue to handle outside of this PR?

justin-stephenson:
@spoore1 This is on my Fedora 29 laptop, let me do more debugging on my end and get back to you.

I created ticket https://github.com/Scribery/tlog/issues/239 for the subprocess module consideration:

PR #237, opened 2019-03-20
journal_json_writer: Remove 'augment' boolean assertion

justin-stephenson:
Don't assert boolean value of `augment`, it is a configurable option
that could be true or false.

I believe if recording or username are NULL and we reach these assert lines,
then the assert won't trigger as expected.

### DIFF
--- a/include/tlog/journal_json_writer.h
+++ b/include/tlog/journal_json_writer.h
@@ -68,9 +68,9 @@ tlog_journal_json_writer_create(struct tlog_json_writer **pwriter,
 {
     assert(pwriter != NULL);
     assert(tlog_syslog_priority_is_valid(priority));
-    assert(!augment || recording != NULL);
-    assert(!augment || username != NULL);
-    assert(!augment || session_id != 0);
+    assert(recording != NULL);
+    assert(username != NULL);
+    assert(session_id != 0);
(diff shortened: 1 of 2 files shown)


include/tlog/journal_json_writer.h
```diff
@@ -68,9 +68,9 @@ tlog_journal_json_writer_create(struct tlog_json_writer **pwriter,
 {
     assert(pwriter != NULL);
     assert(tlog_syslog_priority_is_valid(priority));
-    assert(!augment || recording != NULL);
-    assert(!augment || username != NULL);
-    assert(!augment || session_id != 0);
```
spbnick: These `assert`s basically verify that the caller either didn't ask for augmentation, or if they asked, that they provided the required data to put into the extra fields. This has only to do with verifying the caller is sane. It has nothing to do with whether `augment` is configurable or not. This code doesn't know anything about that and it should not. It should only take care that the parameters it's provided make sense. It will then be the job of the configuration-handling code to verify that no conflicting configuration is possible, which would trip these `assert`s, or, if they're disabled, would make the code crash (because an argument was NULL). In short, separation of concern.

justin-stephenson: Thanks for the explanation! I will close this one out.

PR #243, opened 2019-04-22
Add tlitest module for tlog play tests

spoore1:
New pytest module test_tlog_play.py provides test cases for tlog-play
command.  This includes two test classes covering basic usage and
playback control usage.

### DIFF
--- a/lib/tlitest/misc.py
+++ b/lib/tlitest/misc.py
@@ -136,6 +136,7 @@ def mkrecording(shell, filename=None, sleep=3):
         opts = '-w journal'
     else:
         opts = '-o {}'.format(filename)
+    shell.sendline('stty speed 50')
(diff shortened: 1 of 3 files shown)


justin-stephenson:
@spoore1 Thanks for the changes we discussed, I see `tlitestlocaladmin1` is being added to the systemd-journal group in the setup but the **test_play_from_journal** test is still being run at `tlitestlocaluser1`, could you fix this to allow the test user to read/playback the recording from the journal for this test? I would also add a comment in the setup that mentions RHEL8 requires the user to be added to this group to read the journal.

Otherwise just please remove the small comments below and this one should be good to go:
```
          # pytest.set_trace()

          # shell.logfile = sys.stdout
```

Thanks!

PR #244, opened 2019-04-22
Add tlitest module for tlog-play remote tests

spoore1:
New pytest module test_tlog_play_remote.py provides test cases for
playback of remotely recorded journal entries.  To accomplish this we
also provide journal export files that are imported by the test setup
code in the module.  This requires the use of the systemd-journal-remote
command.  Should already be provided by tlitest-setup in src/tlitest.


justin-stephenson:
I do not see the `systemd-journal-remote` package in the src/tlitest/setup script, could you please add it?

Once I installed this rpm the tests are succeeding for me on Fedora and RHEL, ack from me once the above comment is addressed.

PR #246, opened 2019-04-22
Add tlitest module for tlog-rec performance opts

spoore1:
New pytest module test_tlog_rec_perf_opts.py provides test cases for
the performance related options for the tlog-rec command.   These
include the latency, payload, and limit options.  The test cases also
simulate fast i/o with no delay loops and large files.

### DIFF
--- a/lib/tlitest/test_tlog_rec_perf_opts.py
+++ b/lib/tlitest/test_tlog_rec_perf_opts.py
@@ -1,0 +1,120 @@
+""" tlog tests """
+import os
+import stat
+from tempfile import mkdtemp
+
+from misc import ssh_pexpect, \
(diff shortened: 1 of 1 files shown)


justin-stephenson:
The test `test_record_fast_input_with_limit_action_delay` is failing consistently on Fedora 29, the recording is not logging the last 20 or so test_### numbers. I will need to look into this to determine if there is a bug in the rate limiting code.

justin-stephenson:
@spoore1 The updated delay test has been running successfully for me this morning, I will do more testing on this and get back to you once we are confident that there are no failures.

justin-stephenson:
These tests have been running successfully, merging.

PR #252, opened 2019-05-29
tlitest tlog_rec fix journal_find_last failures

spoore1:
journal_find_last was using _COMM only to match journal entries.  This
can fail to find some entries if they do not include that field.  The
workaround here is to add an additional check for SYSLOG_IDENTIFIER if
_COMM is not found.


lib/tlitest/test_tlog_rec.py
```diff
@@ -46,7 +46,7 @@ def test_record_command_to_journal(self):
         shell = ssh_pexpect(self.user1, 'Secret123', 'localhost')
         reccmd = 'test_record_to_journal'
         shell.sendline('tlog-rec -w journal echo {}'.format(reccmd))
-        find_journal_entry_with_match(reccmd, match_filter, 10)
```
justin-stephenson: Should there be a call to `check_journal` here? I believe that the higher level function which calls `journal_find_last`

lib/tlitest/test_tlog_rec.py
```diff
@@ -59,7 +59,7 @@ def test_record_command_to_syslog(self):
         shell = ssh_pexpect(self.user1, 'Secret123', 'localhost')
         reccmd = 'test_record_to_syslog'
         shell.sendline('tlog-rec --writer=syslog echo {}'.format(reccmd))
-        find_journal_entry_with_match(reccmd, match_filter, 10)
```
justin-stephenson: And here as well

lib/tlitest/test_tlog_rec.py
spoore1: yes, correcting now.

lib/tlitest/test_tlog_rec.py
spoore1: same as above.

Issue #219, opened 2018-10-02
allow tlog-play to send output to a file


spoore1:
Currently tlog-play requires a terminal connected to the output in order to display the output.  This prevents using normal redirects from sending the playback output to a file.   There are cases where this could be very useful in cases like testing and generating other reports.  

From my point of view either a redirect or an output file option would be enough to provide this feature.  Something like:

tlog-play -i /tmp/recorded_session.tlog > /tmp/recorded_session.out
or
tlog-play -i /tmp/recorded_session.tlog -o /tmp/recorded_session.out

Thanks

justin-stephenson:
This is merged and will be fixed in **v7**

PR #253, opened 2019-06-20
Add tlog-play "output" option

justin-stephenson:
Allow redirecting `tlog-play` output to a file with a new `-o`/`--output` option

This is in reference to https://github.com/Scribery/tlog/issues/219

@spbnick @spoore1 Would you mind a quick review?

### DIFF
--- a/lib/tlitest/test_tlog_play.py
+++ b/lib/tlitest/test_tlog_play.py
@@ -51,6 +51,32 @@ class TestTlogPlay:
         out = shell2.expect([pexpect.TIMEOUT, 'KNOWN BUGS'], timeout=10)
         assert out == 1
 
+    def test_play_output_file(self):
+        """
+        Ensure tlog-play will redirect output to file
+        """
+
+        outputfile = mklogfile(self.tempdir)
+        recordedtext = 'output_test'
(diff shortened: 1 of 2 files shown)


spbnick:
I would say this is a bit too involved and adds expands the interface unnecessarily. Simply detect if stdout is a terminal or not (e.g. by checking errno after the failed `tcgetattr`, or with `isatty`), handle that appropriately, and let people use shell redirections, so that `tlog-play >file.bin` works.

lib/tlitest/test_tlog_play.py
```diff
@@ -51,6 +51,32 @@ def test_play_from_journal(self):
...
+
+        playcmd = f'tlog-play -r journal -M {tlog_rec} > {outputfile}'
+        shell2 = ssh_pexpect(self.user1, 'Secret123', 'localhost')
+        shell2.sendline(playcmd)
+        time.sleep(5)
+
+        with open(outputfile) as f:
+            read_data = f.read()
+            assert recordedtext in read_data
+        f.closed
```
spoore1: Would misc.check_outfile() be better here? or don't want/need that here specifically?

PR #258, opened 2019-07-16
Example usage tlog in production

metanovii:
For this example i use ELK stack:
elasticsearch 7.1.1(cluster) + logstash + kibana + nginx, all on each nodes
elasticsearch port 9200 behind nginx proxy with basic auth and ip access control (may be should use https via nginx :)  )
All nodes behind haproxy.

Build packages for ubuntu 16/18 use:  cd build && make packages   :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ Makefile
 Makefile.in
 __pycache__/
 *.py[co]
+*.deb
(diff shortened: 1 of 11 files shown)


justin-stephenson:
Thank you @metanovii ! I would like to test this configuration, do you mind making the Commit Subject/Message a bit more clear?

PR #259, opened 2019-07-16
Add -i/--interactive option to tlog-rec-session

justin-stephenson:
Shells such as **bash** and **tcsh** accept the `-i` option to start an interactive shell.

Pass the `-i` argument to the shell when **tlog-rec-session** is called with
the `-i` argument. This may be useful for applications/programs written
to execute shell programs with `-i`, such as Cockpit.

### DIFF
--- a/lib/tlog/rec_session_conf.c
+++ b/lib/tlog/rec_session_conf.c
@@ -289,6 +289,8 @@ tlog_rec_session_conf_get_shell(struct tlog_errs **perrs,
     struct json_object *args;
     bool login;
     bool command;
+    bool interactive;
+    bool option;
     const char *path;
     char *name = NULL;
     char *buf = NULL;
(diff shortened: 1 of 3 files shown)


lib/tlog/rec_session_conf.c
```diff
@@ -342,9 +348,15 @@ tlog_rec_session_conf_get_shell(struct tlog_errs **perrs,
         grc = TLOG_RC_FAILURE;
         TLOG_ERRS_RAISES("Command string is not specified");
     }
+    if (command && interactive) {
+        grc = TLOG_RC_FAILURE;
+        TLOG_ERRS_RAISES("Specified arguments cannot be used together");
+    }
```
spbnick: I tried various shells, and it seems that both '-c' and '-i' can be used together, so tlog shouldn't put these restrictions on its interface.

justin-stephenson: Thanks Nikolai, I will fix it

PR #261, opened 2019-07-29
ES Reader: Handle missing fields when reading from ES

justin-stephenson:
When playing back from elasticsearch, add empty placeholder value for
fields`(in_txt, host, user)` if these fields are missing in ES database.


justin-stephenson:
Thank you Nikolai, I made those changes. The `schema.json` already excludes **in_txt** and **out_txt** from the required properties list.

Issue #262, opened 2019-08-05
Windows 10 (Ubuntu): Failed retrieving session ID


bmayerornl:
I am using Windows 10 Enterprise 1803 17134.885 with Ubuntu under WSL.

I cloned and built from git and ran with the standard example to write from file:
`tlog-rec --writer=file --file-path=phi.log`

The only output was:
`Failure`
`Failed retrieving session ID`

justin-stephenson:
@bmayerornl This error occurs when tlog is unable to read the contents of file `/proc/self/sessionid` from the filesystem.

spbnick:
In theory, tlog-rec can be modified to not record a session ID (which seems
unavailable on Windows). However, the question is how important it would be
to support this case.

bmayerornl:
It does not look like their is a way to turn that off on the command line?

spbnick:
@bmayerornl, yeah, that's hard-coded logic at the moment, and there's no way to change it from the command line.

shellster:
I also ran into this.  Was able to quickly patch it for my uses, will try to submit a PR when I get a chance.  The quick workaround is to edit `lib/tlog/session.c`, in the get_session_id method.  Replace the logic that is currently trying to get the session id via `/proc/self/sessionid` (this doesn't exist in WSL).  You have to include `#include <unistd.h>` then you can get the session id by calling: `pid = (unsigned int) getsid(0);` and checking that the value doesn't come back as `-1`.  This works in my testing.

shellster:
@spbnick When you get a chance, can you please review my PR and merge if there are no issues?  My testing indicates that this fixes issues, and I don't foresee it breaking anything.  I believe `getsid` is more more standard and reliable, anyway.

VincentFortuneDeng:
> I also ran into this. Was able to quickly patch it for my uses, will try to submit a PR when I get a chance. The quick workaround is to edit `lib/tlog/session.c`, in the get_session_id method. Replace the logic that is currently trying to get the session id via `/proc/self/sessionid` (this doesn't exist in WSL). You have to include `#include <unistd.h>` then you can get the session id by calling: `*pid = (unsigned int) getsid(0);` and checking that the value doesn't come back as `-1`. This works in my testing.

我也遇到了同样的问题，终于找到解决方案了；感谢~~

PR #264, opened 2019-08-17
Updated get_session_id to use getsid to be compatible with WSL

shellster:
Addresses issue https://github.com/Scribery/tlog/issues/262

### DIFF
--- a/lib/tlog/session.c
+++ b/lib/tlog/session.c
@@ -28,31 +28,21 @@
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <unistd.h>
 
 tlog_grc
 tlog_session_get_id(unsigned int *pid)
-{
-    FILE *file;
-    int orig_errno;
-    int rc;
-
+{  
     assert(pid != NULL);
 
-    file = fopen("/proc/self/sessionid", "r");
-    if (file == NULL) {
+    *pid = (unsigned int) getsid(0);
+    
+    if (*pid == -1) {
         return TLOG_RC_FAILURE;
     }
-    rc = fscanf(file, "%u", pid);
-    orig_errno = errno;
-    fclose(file);
-    if (rc == 1) {
+    else {
         return TLOG_RC_OK;
     }
-    if (rc == 0) {
-        return TLOG_GRC_FROM(errno, EINVAL);
-    }
-
-    return TLOG_GRC_FROM(errno, orig_errno);
 }
 
 /**
--- a/lib/tlog/session.c
+++ b/lib/tlog/session.c
@@ -35,12 +35,13 @@ tlog_session_get_id(unsigned int *pid)
 {  
     assert(pid != NULL);
 
-    *pid = (unsigned int) getsid(0);
+    pid_t session = getsid(0);
     
-    if (*pid == -1) {
+    if (session == -1) {
         return TLOG_RC_FAILURE;
     }
     else {
+	*pid = (unsigned int) session;
         return TLOG_RC_OK;
     }
 }


spbnick:
Thank you for the patch, @shellster.

However, `getsid(2)` returns ID of a different session from the one stored in `/proc/self/sessionid`. The former is a process group management object which can be created by anything, and the latter is an audit user session object, created only when a user logs in and destroyed when the user logs out. Tlog-rec-session needs the latter so the administrator or the integrating system is able to correlate the recording with audit events belonging to the same session. Tlog-rec, strictly-speaking, doesn't need it. However we need to evaluate which impact it would have on our plans to have it removed from tlog-rec and how exactly to best do it.

I'm closing this PR, as we can't merge it in this shape. Thank you.

shellster:
Thanks for responding to my pull.  Seems I fundamentally misunderstood what the sessionid was trying to grab.  However, upon further research, it looks like this concept of sessionid doesn't exist in WSL, and possibly doesn't exist in other non-systemd configurations. `getsid` is returning the pid of the parent-most shell process, effectively. Would an acceptable solution be to leave the old code, but fall back to `getsid` if `/proc/self/sessionid` does not exist? That way the code would work as expected in most situations, and still provide a useful session id on environments that don't offer `/proc/self/sessionid`.  I'm happy to update my PR if you would consider this.  My ultimate goal is to have a work version of tlog that I can use across all my environments, and I'd prefer to get something into the master branch as opposed to my own build.

spbnick:
The sessionid functionality is not part of systemd, but rather the [audit subsystem in the Linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/audit.c?id=d1abaeb3be7b5fa6d7a1fbbd2e14e3310005c4c1#n2236).

The ID returned by `getsid()` wouldn't really be useful, I think. Do you have anything specific in mind? Regardless, we won't change semantics of the `session` field, if we're to use `getsid()` ID, then it would need to be a separate field. Perhaps using the /proc/self/sessionid could be a build-time (`configure`) option. We would then need to designate the `session` JSON field as optional.

shellster:
Either way, this doesn't appear to exist in WSL.  If you don't want to use `getsid()`, can we just set the `sessionid` to `-1` instead of causing a hard error? My goal, again, is to be able to use tlog in WSL as well as all my linux systems and right now that isn't possible.  I get the desire not to pollute the meaning of `sessionid`.  I also think it is important not to get rid of the field for systems that have it.  Removing it on systems that don't can cause parsing ambiguity which is also not ideal. Therefore, maybe the best plan is just to set it to a non-valid value of `-1`.  Obviously, this would require reworking the value to be signed instead of unsigned.  I could submit a patch for that.

spbnick:
@shellster, do you want to use the same binary both on Linux and on WSL?

shellster:
Doesn't have to be the same binary, but I think it would be great if I can build it from the same source as opposed to building and maintaining an entire separate fork of the project. I don't really care what your preferred solution is as far as the nature of the edit, I would just like to see it implemented sooner than later in the main repo, and I'm happy to help by submitting a patch if that's helpful. From my perspective, that sessionid field appears to serve a purpose on most versions of *nix.  So, I don't think removing it makes sense.  I think not having this field on WSL systems and other ones that don't support all of "audit" is also not a good idea because of the JSON parsing ambiguity in the logs for those systems.  But I'll that is obviously your decision.

spbnick:
I see. OK.

First of all, I think omitting the "session" field from JSON when the session ID is unavailable is a cleaner solution than using the invalid value. For several reasons: if we admit an signed type, then there will be more than one invalid value which could lead to bugs; it would be easier to discover non-complying code, because a missing field would more likely result in a clear fault, than a negative value, which could simply be read as e.g. UINT_MAX; finally, a missing field logically corresponds more directly to a missing value than an invalid value.

I'm a little wary about simply assuming the system doesn't provide audit session IDs at all, if we just fail to open the `/proc/self/sessionid` file, but I think it's a good-enough assumption which we can start with. So, we can start with detecting ENOENT on opening the `/proc/self/sessionid` file, assuming the system doesn't provide the session ID, passing the invalid ID (-1) to the JSON writer and omitting the field from JSON output in this case.

Also, this change needs to be accompanied by a verification that the tlog code is also using -1 as an invalid session ID, because it could be using zero, which would be technically wrong, as the kernel can return audit session ID zero on roll-over of its atomic_t counter. If that's the case, the tlog code needs to be switched to using -1 as well.

Since I'm no longer closely working on the project, I would leave this to you @shellster, or to @justin-stephenson - the new developer.

shellster:
Thanks for your response.  I think everything you said makes sense:

> I'm a little wary about simply assuming the system doesn't provide audit session IDs at all, if we just fail to open the /proc/self/sessionid file, but I think it's a good-enough assumption which we can start with.  

As other corner cases arise (for instance systems that require the audit session id to be derived in different ways, they can be handled in this same method.

Can you clarify this paragraph?

> Also, this change needs to be accompanied by a verification that the tlog code is also using -1 as an invalid session ID, because it could be using zero, which would be technically wrong, as the kernel can return audit session ID zero on roll-over of its atomic_t counter. If that's the case, the tlog code needs to be switched to using -1 as well.

I understand that it is possible for the session id to be `0`, so we want to make sure that the code consistently uses a signed int with a `-1` for invalid/missing session id.  Are you saying that there may be other places in the code that are incorrectly passing a `0` right now? Also do we know what the data type is for the audit session id?  Will want to verify that using whatever signed value we use is big enough to handle all valid session id's.

justin-stephenson:
> Can you clarify this paragraph?
> 
> > Also, this change needs to be accompanied by a verification that the tlog code is also using -1 as an invalid session ID, because it could be using zero, which would be technically wrong, as the kernel can return audit session ID zero on roll-over of its atomic_t counter. If that's the case, the tlog code needs to be switched to using -1 as well.
> 
> I understand that it is possible for the session id to be `0`, so we want to make sure that the code consistently uses a signed int with a `-1` for invalid/missing session id. Are you saying that there may be other places in the code that are incorrectly passing a `0` right now? Also do we know what the data type is for the audit session id? Will want to verify that using whatever signed value we use is big enough to handle all valid session id's.

I read Nikolai's comment as once we switch to using a normal signed integer data type, then we will also want to consider session == 0 as an invalid ID, converting it to -1 then the code added to handle the invalid ID will be executed in either case.

Also, if the session field is omitted then tlog_json_msg_init() will need to be modified to allow this field as optional or tlog-play will fail with TLOG_RC_JSON_MSG_FIELD_MISSING.

spbnick:
Yes, I suspect that tlog might be using zero as invalid session ID right now. I just don't remember what I did there. However, after looking at the kernel code for a while, I realize that it uses an `unsigned int` for the session ID and `(unsigned int)-1` as the invalid value, which translates to UINT_MAX. So that's what we should use, not a signed `-1`.

Technically, though, it might be difficult to determine the actual kernel's UINT_MAX, as the kernel might be 64-bit and the userspace (and tlog) 32-bit. This might need additional research to figure out.

PR #266, opened 2019-10-15
Modernize packaging and add Debian/Ubuntu support

Conan-Kudo:
This pull request modernizes the packaging spec and also trivially adds support for building for Debian/Ubuntu natively using [debbuild](https://github.com/debbuild/debbuild).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
 - $CLANG_SCAN_BUILD make
 - $CLANG_SCAN_BUILD make check
 - make distcheck
-- rpmbuild --define "_tmpfilesdir /usr/lib/tmpfiles.d" -ta --nodeps tlog-*.tar.gz
+- rpmbuild -ta --nodeps tlog-*.tar.gz
 - test "x$DISTCHECK" = xyes && make distcheck || true
 after_success:
 - test "x$COVERAGE" = xyes -a "x$CC" = "xgcc" && lcov --compat-libtool --directory ./lib/tlog/ --capture --output-file coverage.info && coveralls-lcov coverage.info
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 sudo: required
-dist: trusty
+dist: xenial
 env:
 - CONFIGURE_ARGS="" DISTCHECK=yes
 - CONFIGURE_ARGS="--enable-debug" DISTCHECK=yes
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@justin-stephenson ping? Could you please take a look at this pull request?

justin-stephenson:
Hello @Conan-Kudo Thanks for the pull request! I do agree the spec file has some outdated syntax and content, that is something I will work on fixing soon.

I am hesitant to add the debbuild lines as the spec file is intended for use with 'rpm' on Fedora/Red Hat distribution and the file needs to conform to Fedora Packaging guidelines.

Is there any problem building tlog from source on Ubuntu/Debian currently? I don't believe tlog is added into debian/ubuntu repositories, last I remember.

Conan-Kudo:
> Is there any problem building tlog from source on Ubuntu/Debian currently? I don't believe tlog is added into debian/ubuntu repositories, last I remember.

It doesn't exist in Debian/Ubuntu, and at Datto, we want to use this on our Ubuntu systems. We generally package for Debian/Ubuntu using debbuild and since this project has a spec file in the source tree, I wanted to contribute our improvements back.

The spec should generally conform to the Fedora Packaging Guidelines as it stands now. If you wanted to remove the Debian/Ubuntu conditionals when using the spec downstream in Fedora, it should be very minimal effort to strip. But I'm hoping you'd accept this and leverage this to help support builds for Debian/Ubuntu users in the future...

justin-stephenson:
Thank you for clarification.

As stated in Fedora Packaging Guidelines:

> To help facilitate legibility, only macros and conditionals for Fedora and EPEL are allowed to be used in Fedora Packages. Use of macros and conditionals for other distributions, including Fedora derivatives, is not permitted in spec files of packages in the main Fedora repositories unless those macros and conditionals are also present in Fedora.

I would like to avoid stripping Debian/Ubuntu conditionals from the spec file when packaging tlog for Fedora. I would like to help support users of tlog on Debian/Ubuntu if possible, I propose requesting tlog to be added into Debian/Ubuntu official repositories is the optimal path for this. In that case package mantainers can make distribution requisite build changes as necessary. I also did not see any mention of `debbuild` in official Debian packaging documentation.

Conan-Kudo:
debbuild is a relatively new method of producing Debian packages, and does not currently exist in Debian itself.

Insofar as the FPG rule about conditionals, it is not generally enforced, as you can see from these examples:
* https://src.fedoraproject.org/rpms/ceph/blob/master/f/ceph.spec
* https://src.fedoraproject.org/rpms/abrt/blob/master/f/abrt.spec
* https://src.fedoraproject.org/rpms/mock/blob/master/f/mock.spec
* https://src.fedoraproject.org/rpms/mock-core-configs/blob/master/f/mock-core-configs.spec

debbuild itself is also packaged and available in Fedora: https://apps.fedoraproject.org/packages/debbuild

Conan-Kudo:
I'm not planning on writing dsc based Debian packaging. If you don't want the debbuild change, I'll yank the commit.

justin-stephenson:
I am happy to merge the updated spec file into `debian/tlog.spec` if it will be useful to others.

If someone later wants to add dsc based Debian packaging files then they can be added under this subdirectory also.

Conan-Kudo:
@justin-stephenson I'm not sure it makes sense to do that. Part of the reason why I did it this way is that I expect it will be easily kept up to date, since the rpm packaging work would just be automatically inherited for producing the deb.

justin-stephenson:
@Conan-Kudo Sorry I was out all week on PTO, okay I will review the proposed changes and merge after doing some testing on my end.

justin-stephenson:
Thanks for the contribution @Conan-Kudo !

Conan-Kudo:
@justin-stephenson Thanks for accepting it!

Conan-Kudo:
@justin-stephenson Do you have plans to cut a new release of tlog anytime soon?

justin-stephenson:
@Conan-Kudo I plan on doing a new release in the next 1 or 2 weeks.

PR #270, opened 2019-11-25
RHEL6 rpmbuild Spec file fixes

justin-stephenson:


tlog.spec
```diff
@@ -2,9 +2,9 @@
 
 %if 0%{?rhel} == 0 && 0%{?rhel} < 7
 # If it's RHEL6 and older
-%bcond_with systemd
-%else
 %bcond_without systemd
+%else
+%bcond_with systemd
```
Conan-Kudo: This inverts the logic and makes it so that `systemd` dependencies are enabled on RHEL 6 and older while being disabled everywhere else.

tlog.spec
```diff
@@ -2,9 +2,9 @@
 
 %if 0%{?rhel} == 0 && 0%{?rhel} < 7
```
Conan-Kudo: I see the mistake now. This needs to be `%if 0%{?rhel} && 0%{?rhel} < 7`...

PR #271, opened 2019-11-26
Fix spec file configure conditional

justin-stephenson:


tlog.spec
```diff
@@ -75,7 +75,7 @@ in JSON format.
 %setup -q
 
 %build
-%configure --disable-rpath --disable-static %{!?with_systemd:--disable-journal}
+%configure --disable-rpath --disable-static %{!?with systemd:--disable-journal}
```
Conan-Kudo: I'm pretty sure this was only broken because of PR #270.

PR #272, opened 2019-12-02
Fix spec file conditionals to build properly on EL6 and add missing BR for Debian/Ubuntu

Conan-Kudo:
The conditions were accidentally set to force all distributions
to invoke the systemd dependencies, instead of allowing RHEL/CentOS 6
to not have systemd dependencies.

Also, add missing `BuildRequires: pkg-config` for Debian/Ubuntu builds.

Fixes #270 and #271.

### DIFF
--- a/tlog.spec
+++ b/tlog.spec
@@ -1,10 +1,10 @@
 %global _hardened_build 1
 
-%if 0%{?rhel} == 0 && 0%{?rhel} < 7
+%if 0%{?rhel} && 0%{?rhel} < 7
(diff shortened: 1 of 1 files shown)


justin-stephenson:
Thanks for the PR @Conan-Kudo - let me test out the changes and respond back once done.

justin-stephenson:
This fixes the problems we were seeing on RHEL8 builds, I understand now that %bcond has inverted logic. I will quickly test on RHEL6, and RHEL7 and then merge if everything looks good.

Issue #227, opened 2018-12-10
tlog-rec-session breaks Ansible pipelining


mikedep333:
Test setup the Ansible-manged system, ldt-4009653:
- RHEL 7.6 w/ tlog v5 (almost fresh install)
- local account "admin-renamed" with sudo privileges
- admin-renamed has its shell set to /usr/bin/tlog-rec-session
- default tlog-rec & tlog-rec-session settings
- "requiretty" disabled in /etc/sudoers (per Ansible pipelining instructions)

Test setup on my Ansible management system, ldt-4323517:
- RHEL 7.6 & Ansible 2.7 or 2.8 (development branch)
- In ansible.cfg:
```
[ssh_connection]
pipelining = True
```

Running this ansible test command (and entering the password when prompted) fails:
```
[Michael.DePaulo@ldt-4323517 Ansible]$ ansible -vvv -m setup -i ldt-4009653, ldt-4009653 -u admin-renamed -k --become-method sudo --become
ansible 2.8.0.dev0
  config file = /home/Michael.DePaulo/linux-desktop/Ansible/ansible.cfg
  configured module search path = [u'/home/Michael.DePaulo/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.5 (default, Sep 12 2018, 05:31:16) [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
Using /home/Michael.DePaulo/linux-desktop/Ansible/ansible.cfg as config file
SSH password: 
Parsed ldt-4009653, inventory source with host_list plugin
META: ran handlers
Using module file /usr/lib/python2.7/site-packages/ansible/modules/system/setup.py
<ldt-4009653> ESTABLISH SSH CONNECTION FOR USER: admin-renamed
[...]

PR #276, opened 2020-02-24
Handle piped in I/O from STDIN

justin-stephenson:
Modify the main recording loop to transfer I/O until an EOF
is signalled only from the forked recorded process.

Also add functionality to close input or output file descriptors
to the sink interface. This is necessary to close one end of a pipe
when I/O is pushed to an interactive process like python via stdin.

Resolves https://github.com/Scribery/tlog/issues/227

### DIFF
--- a/include/tlog/sink_type.h
+++ b/include/tlog/sink_type.h
@@ -95,6 +95,15 @@ typedef tlog_grc (*tlog_sink_type_cut_fn)(struct tlog_sink *sink);
 typedef tlog_grc (*tlog_sink_type_flush_fn)(struct tlog_sink *sink);
 
 /**
+ * IO Close function prototype.
+ *
+ * @param sink    The sink to close an associated I/O file descriptor.
+ * @param output  Indicates which sink file descriptor to close.
+ */
+typedef void (*tlog_sink_type_io_close_fn)(struct tlog_sink *sink,
+                                           bool output);
+
+/**
  * Cleanup function prototype.
  *
  * @param sink  The sink to cleanup.
@@ -109,6 +118,7 @@ struct tlog_sink_type {
     tlog_sink_type_write_fn     write;      /**< Writing function */
     tlog_sink_type_cut_fn       cut;        /**< I/O-cutting function */
     tlog_sink_type_flush_fn     flush;      /**< Flushing function */
+    tlog_sink_type_io_close_fn  io_close;   /**< I/O-close function */
     tlog_sink_type_cleanup_fn   cleanup;    /**< Cleanup function */
 };
 
--- a/lib/tlog/pkt.c
+++ b/lib/tlog/pkt.c
@@ -66,6 +66,21 @@ tlog_pkt_init_window(struct tlog_pkt *pkt,
 }
 
 void
+tlog_pkt_init_eof(struct tlog_pkt *pkt,
+                  const struct timespec *timestamp,
+                  bool output)
+{
+    assert(pkt != NULL);
+    memset(pkt, 0, sizeof(*pkt));
+    pkt->timestamp = *timestamp;
+    pkt->type = TLOG_PKT_TYPE_IO;
+    pkt->data.io.output = output;
+    pkt->data.io.len = 0;
+    assert(tlog_pkt_is_valid(pkt));
+    assert(tlog_pkt_is_eof(pkt));
+}
+
+void
 tlog_pkt_init_io(struct tlog_pkt *pkt,
                  const struct timespec *timestamp,
                  bool output,
@@ -112,6 +127,13 @@ tlog_pkt_is_void(const struct tlog_pkt *pkt)
 {
     assert(tlog_pkt_is_valid(pkt));
     return pkt->type == TLOG_PKT_TYPE_VOID;
+}
+
+bool
+tlog_pkt_is_eof(const struct tlog_pkt *pkt)
+{
+    assert(tlog_pkt_is_valid(pkt));
+    return pkt->type == TLOG_PKT_TYPE_IO && pkt->data.io.len == 0;
 }
 
 bool
--- a/lib/tlog/rec.c
+++ b/lib/tlog/rec.c
@@ -66,12 +66,21 @@ static volatile sig_atomic_t tlog_rec_alarm_set;
 /* Number of ALRM signals caught */
 static volatile sig_atomic_t tlog_rec_alarm_caught;
 
+/* True if child stopped or terminated */
+static volatile sig_atomic_t tlog_rec_child_exited;
+
 static void
 tlog_rec_alarm_sighandler(int signum)
 {
     (void)signum;
     tlog_rec_alarm_set = false;
     tlog_rec_alarm_caught++;
+}
+
+static void
+tlog_rec_sigchld_handler()
+{
+    tlog_rec_child_exited = true;
 }
 
 /**
@@ -808,7 +817,7 @@ tlog_rec_transfer(struct tlog_errs    **perrs,
 {
     const int exit_sig[] = {SIGINT, SIGTERM, SIGHUP};
     tlog_grc return_grc = TLOG_RC_OK;
-    tlog_grc grc;
+    tlog_grc grc = TLOG_RC_OK;
     size_t i, j;
     struct sigaction sa;
     bool log_pending = false;
@@ -821,6 +830,7 @@ tlog_rec_transfer(struct tlog_errs    **perrs,
     tlog_rec_exit_signum = 0;
     tlog_rec_alarm_set = false;
     tlog_rec_alarm_caught = 0;
+    tlog_rec_child_exited = false;
 
     /* Setup signal handlers to terminate gracefully */
     for (i = 0; i < TLOG_ARRAY_SIZE(exit_sig); i++) {
@@ -856,10 +866,28 @@ tlog_rec_transfer(struct tlog_errs    **perrs,
                           "Failed to set a SIGALRM signal action");
     }
 
+    /* Setup SIGCHLD signal handler */
+    sa.sa_handler = tlog_rec_sigchld_handler;
+    sigemptyset(&sa.sa_mask);
+    /* NOTE: no SA_RESTART on purpose */
+    sa.sa_flags = 0;
+    if(sigaction(SIGCHLD, &sa, NULL) == -1) {
+        grc = TLOG_GRC_ERRNO;
+        TLOG_ERRS_RAISECS(grc,
+                          "Failed to set a SIGCHLD signal action");
+    }
+
     /*
      * Transfer I/O and window changes
      */
     while (tlog_rec_exit_signum == 0) {
+        /* Expected exit conditions */
+        if (tlog_rec_child_exited) {
+            if (grc == TLOG_GRC_FROM(errno, EIO) || (tlog_pkt_is_eof(&pkt))) {
+                break;
+            }
+        }
+
         /* Handle latency limit */
         new_alarm_caught = tlog_rec_alarm_caught;
         if (new_alarm_caught != last_alarm_caught) {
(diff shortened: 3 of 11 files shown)


justin-stephenson:
@spbnick Do you mind giving a final ack for this? I will be adding integration tests before merging to avoid regressions.

justin-stephenson:
@spbnick thank you again for your review. I updated the code based on your comments, I hope it is more closely aligned with your thinking.

One draft of this code waited for both source input and output to be closed before a Void packet is returned from the source(to break out of the transfer loop) but this would hang when no STDIN input was available. for example:

	[jstephen@agalloch tlog]$ ssh recuser@remotehost sh -c 'echo PLATFORM; uname; echo FOUND; command -v /usr/bin/python'

This code appears to handle all the cases I tested of varying input with remote command execution correctly.

lib/tlog/rec.c
```diff
@@ -926,6 +926,8 @@ tlog_rec_transfer(struct tlog_errs    **perrs,
                 return_grc = grc;
             }
             break;
+        } else if (tlog_pkt_is_eof(&pkt)) {
+            tlog_sink_io_close(tty_sink, pkt.data.io.output);
```
spbnick: Reading this feels nice :)

lib/tlog/sink.c
```diff
@@ -135,6 +135,20 @@ tlog_sink_flush(struct tlog_sink *sink)
...
 
+void
+tlog_sink_io_close(struct tlog_sink *sink,
+                bool output)
+{
+    assert(sink == NULL || tlog_sink_is_valid(sink));
+
+    if (sink == NULL) {
+        return;
+    }
```
spbnick: I think the sink cannot be NULL if we want to operate on it. This is not like we're destroying it so we don't care it's already destroyed. So we shouldn't be trying to handle it being NULL, and should do a hard abort when it happens. I.e. remove `sink == NULL` from the `assert` and drop that return condition.

include/tlog/sink_type.h
```diff
@@ -109,6 +118,7 @@ struct tlog_sink_type {
     tlog_sink_type_write_fn     write;      /**< Writing function */
     tlog_sink_type_cut_fn       cut;        /**< I/O-cutting function */
     tlog_sink_type_flush_fn     flush;      /**< Flushing function */
+    tlog_sink_type_close_fn     close;      /**< Close function */
```
spbnick: I think this should be called `io_close` and `I/O-closing function` to match everything else.

lib/tlog/sink.c
```diff
@@ -135,6 +135,20 @@ tlog_sink_flush(struct tlog_sink *sink)
...
+    if (sink->type->cleanup != NULL) {
```
spbnick: You probably meant `sink->type->close != NULL` here. I would say `sink->type->io_close != NULL` would be better still :)

lib/tlog/tty_source.c
```diff
@@ -274,10 +277,13 @@ tlog_tty_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
...
-                                 tty_source->fd_idx ==
-                                    TLOG_TTY_SOURCE_FD_IDX_OUT,
-                                 tty_source->io_buf, false, rc);
+                tlog_pkt_init_io(pkt, &ts, output, tty_source->io_buf, false, rc);
+            } else if (rc == 0) {
+                tty_source->fd_list[tty_source->fd_idx].fd = -1;
+
+                if (tty_source->fd_list[TLOG_TTY_SOURCE_FD_IDX_OUT].fd >= 0) {
+                    tlog_pkt_init_eof(pkt, output);
+                }
```
spbnick: Can you add a comment explaining what's going on here?

lib/tlog/tty_sink.c
```diff
@@ -51,6 +51,20 @@ tlog_tty_sink_is_valid(const struct tlog_sink *sink)
     return tty_sink != NULL;
 }
 
+static void
+tlog_tty_sink_io_close(struct tlog_sink *sink, bool output)
+{
+    struct tlog_tty_sink *tty_sink =
+                                (struct tlog_tty_sink *)sink;
+
+    int *fd = output ? &tty_sink->out_fd : & tty_sink->in_fd;
```
spbnick: There's an extra space after `&` here, which makes it look like the binary AND operation.

lib/tlog/tty_source.c
```diff
@@ -274,10 +277,18 @@ tlog_tty_source_read(struct tlog_source *source, struct tlog_pkt *pkt)
...
+                tlog_pkt_init_io(pkt, &ts, output, tty_source->io_buf, false, rc);
+            } else if (rc == 0) {
+                tty_source->fd_list[tty_source->fd_idx].fd = -1;
+
+                /* Input:  Initialize and return a zero-length packet indicating EOF.
+                 *         No remaining I/O from this file descriptor
+                 * Output: Return a Void packet, end of source
+                 *         data
+                 */
+                if (tty_source->fd_list[TLOG_TTY_SOURCE_FD_IDX_OUT].fd >= 0) {
```
spbnick: I'm still struggling with understanding this. I'll try to talk through it:

So, if the read returns zero data, it means we've got to EOF on that FD. So, we mark that FD as such by assigning an invalid FD value to its index (could be useful to explain that step as well). Then we check that the output FD has not got the EOF yet. That means that we just got an EOF on the input FD. So, first question: why don't we just check that we're handling the input FD? I.e. why don't we write it more directly:

```c
if (!output) {
    tlog_pkt_init_eof(pkt, output);
}
```

Next, I don't understand **why** we have to return the EOF packet for input, but the void packet for output? Could you explain *that*?

justin-stephenson: HI @spbnick no problem, I prefer to wait until we get these concerns addressed before pushing this.

First question: Yes, that is more direct to use the code you pasted. I will update the PR to use that approach.

Second question: We return an EOF packet for input because we still need to transfer and consume output from the recorded process after input read returns zero data. If we return a Void packet for input then we will prematurely break out of the main transfer loop before reading output I/O.

If we wait for input **and** output to close before returning a Void packet, then tlog will hang in instances where no input data is available because a poll event is never generated for input, and thus we never reach the code to 'close' the input I/O File descriptors and invalidate the input FD index.

Therefore the logic that has worked in my testing is to continue transferring I/O after a read returns 0 for input, but return a Void packet to finish and break out of transferring I/O after output read returns 0. This may be problematic if both input and output I/O is available and the  TTY source poll() triggers generates an output FD event **before** an input event because then we will miss the input I/O entirely. This did not occur in my testing, or with the ansible pipelining case so I'm not sure how to best handle this case.

Hope this helps.

spbnick: Thank you for the explanations, @justin-stephenson. However, I think we strayed off path here quite a bit.

Take a look at the two programs below.

`shut_and_wait`:
```bash
#!/bin/bash

echo "Shutting up!"
# Close all I/O
exec 0<&- 1>&- 2>&-
# Shouldn't block
cat
# Shouldn't print anything
echo "Blah"
echo "Blah" >&2
sleep 5
```

`ssh_shut_up`:
```bash
#!/bin/bash

# Close all I/O
exec 0<&- 1>&- 2>&-
ssh localhost ./shut_and_wait
```
If you put the first one (`shut_and_wait`) into your home folder, and then execute `ssh localhost ./shut_and_wait`, you will note that it only prints the `Shutting up!` message, and doesn't print anything else, nor does it block on `cat`, because all inputs and outputs are closed, and yet `ssh` doesn't finish until the program exits after 5 seconds. This proves that `ssh` doesn't exit until the child exits and no I/O conditions matter. Moreover, if you execute `ssh_shut_up` you will see that it also waits 5 seconds, even though all the I/O file descriptors on the **host** are closed too.

So, it's good that we're fixing the `tty_source` to not return the void packet on any files being closed and not making tlog exit, but we should be going further and not exiting until the child exits, no matter which file descriptors are closed. However, we should still track which FDs are closing and close the corresponding end to communicate properly.

So, I think `tty_source` should simply be returing an EOF packet for **either** input and output EOF, and only ever return the void packet if it cannot return any data at all, i.e. when **all** FDs are closed and `win_fd` is invalid (that is we're not reporting window resizes). At the same time, `tlog_rec_transfer` shouldn't be exiting even if it gets the void packet from the `tty_source`, but only if it detects that the child has exited. I'll leave it to you to research how to best do the latter.

justin-stephenson: PR Updated based on your comments here, hopefully this version looks better. 

One comment that might be worth mentioning:
- Exiting the `tlog_rec_transfer` loop immediately when **SIGCHLD** is caught results in incomplete I/O transferred(including with SA_RESTART), therefore I use the `read_pending` conditional, separate **SIGCHLD** handler, and the below code to complete transferring all I/O before exiting.
~~~
+        /* Exit the loop */
+        if (tlog_rec_child_exited && !read_pending) {
+            break;
+        }
~~~

Thank you again for the insight and review Nikolai.

spbnick: Right, I think we should go even further than that. Please see my review comments.

lib/tlog/rec.c
```diff
@@ -856,10 +867,26 @@ tlog_rec_transfer(struct tlog_errs    **perrs,
...
+    }
+
     /*
      * Transfer I/O and window changes
      */
     while (tlog_rec_exit_signum == 0) {
+        /* Exit the loop */
+        if (tlog_rec_child_exited && !read_pending) {
+            break;
+        }
```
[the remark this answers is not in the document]
justin-stephenson: Wouldn't it need to be `tlog_pkt_pos_is_past(&tty_pos, &pkt)` to ensure that the tty_pos is **at**/**equal to** the same position as the packet ?

I'm having some trouble implementing this, specifically how to determine the end of process output to handle both cases:
~~~
tlog-rec-session -c ls

ssh recuser@remotehost sh -c 'echo PLATFORM; uname; echo FOUND; command -v /usr/bin/python'
~~~

The first command will eventually return errno EIO from the source, I have been successful with adding this code after the `/* Read new data */` block in tlog_rec_transfer:
~~~
        if (tlog_rec_child_exited && (grc == TLOG_GRC_FROM(errno, EIO)) && tlog_pkt_pos_is_past(&tty_pos, &pkt)) {
            break;
        }
~~~
This addresses the issue of logging and delivering complete process output from the recorded process but I am struggling with the end condition of the pipe case. Should we break when we read an zero-byte source 'output' EOF packet ? What would be the best way to ensure all output is consumed before exiting the loop? Any suggestions are appreciated.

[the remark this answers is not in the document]
spbnick: Yeah, `tlog_pkt_pos_is_past(&tty_pos, &pkt)` would work better, since you won't have to negate the return value :)

Yeah, hanging sounds bad, sorry about that, and the problem is probably in termination/transfer logic. If you want to rule out any JSON encoding issues, try to disable logging everything (i.e. input, output, and window changes) and see if it still behaves the same. I suspect it will. I could also recommend backing off transfer logic changes one by one and seeing what makes it hang.

I have a feeling that our main transfer loop should exit after the child has finished, and we've transferred everything from it. See if you can make that condition clearer in the code, and make sure you understand all the termination conditions together. Perhaps something will pop out at you.

justin-stephenson:
Review Items should all be addressed now, this latest code addresses the ansible pipelining problem and passes all integration tests, and variations of input testing. All output is transferred when testing against a program that quickly generates output and exits.

lib/tlog/pkt.c
```diff
@@ -65,6 +65,19 @@ tlog_pkt_init_window(struct tlog_pkt *pkt,
...
+tlog_pkt_init_eof(struct tlog_pkt *pkt,
+                  bool output)
+{
+    assert(pkt != NULL);
+    memset(pkt, 0, sizeof(*pkt));
+    pkt->type = TLOG_PKT_TYPE_IO;
+    pkt->data.io.output = output;
+    pkt->data.io.len = 0;
+    assert(tlog_pkt_is_valid(pkt));
+    assert(tlog_pkt_is_void(pkt));
```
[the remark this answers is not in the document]
justin-stephenson: This is now fixed.

spbnick:
Glad to hear that!

Issue #281, opened 2020-03-19
Omit non-required fields in JSON message if value cannot be determined


justin-stephenson:
In https://github.com/Scribery/tlog/issues/269 tlog was unable to determine a valid hostname, it may be best to exclude the hostname JSON property from the message in this case. Similar logic should be applied to other non-required message fields.

PR #282, opened 2020-03-19
Use empty string on hostname resolution failure

justin-stephenson:
Fallback to using an empty string as the JSON hostname field when the system hostname is unable to be determined, instead of erroring out.

A more complete fix is separated out and should be handled in https://github.com/Scribery/tlog/issues/281


spbnick:
LGTM, but I wonder if we should print a message when this happens to help administrators.

PR #284, opened 2020-03-25
issue 283: tlog-play improve auth options. 

ajf8:
As well as being able to specify authentication in the URL, provide some other
options for configuring the curl client that connects to the ElasticSearch API.

* user:pass in the TLOG_CURL_USERPWD environment variable
* In a ~/.netrc file
* Enable GSSNEGOTIATE for Kerberos support.

For kerberos the username must be non-empty, either in the URL or $TLOG_CURL_USERPWD.

### DIFF
--- a/lib/tlog/es_json_reader.c
+++ b/lib/tlog/es_json_reader.c
@@ -210,6 +210,7 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
     const char *base_url = va_arg(ap, const char *);
     const char *query = va_arg(ap, const char *);
     size_t size = va_arg(ap, size_t);
+    bool verbose = (bool)va_arg(ap, int);
     CURLcode rc;
     tlog_grc grc;
 
@@ -229,6 +230,47 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
         grc = TLOG_GRC_FROM(curl, rc);
         goto error;
     }
+
+    /* Allow kerberos (negotiate) authentication */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
+    if (rc != CURLE_OK)
+        if (rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
+
+    /* Read credentials from .netrc if present */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
+    if (rc != CURLE_OK)
+        if (rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
+
+    /* On >= 400 responses fail immediately, before trying to parse JSON */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_FAILONERROR, 1);
+    if (rc != CURLE_OK)
+        if (rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
+
+    /* Take credentials from TLOG_CURL_USERPWD environment variable if present */
+    if (getenv("TLOG_CURL_USERPWD")) {
+        rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_USERPWD, getenv("TLOG_CURL_USERPWD"));
+        if (rc != CURLE_OK) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
+    }
+
+    /* Configure verbose logging on the curl client. */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_VERBOSE, verbose);
+    if (rc != CURLE_OK)
+        if (rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
 
     /* Format URL prefix */
     grc = tlog_es_json_reader_format_url_pfx(&es_json_reader->url_pfx,
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -82,6 +82,7 @@ tlog_play_create_es_json_reader(struct tlog_errs **perrs,
     tlog_grc grc;
     const char *baseurl;
     const char *query;
+    bool verbose;
     struct tlog_json_reader *reader = NULL;
     struct json_object *obj;
 
@@ -94,6 +95,10 @@ tlog_play_create_es_json_reader(struct tlog_errs **perrs,
         TLOG_ERRS_RAISES("Elasticsearch base URL is not specified");
     }
     baseurl = json_object_get_string(obj);
+
+    /* Check for the verbose flag */
+    verbose = json_object_object_get_ex(conf, "verbose", &obj) &&
+        json_object_get_boolean(obj);
 
     /* Check base URL validity */
     if (!tlog_es_json_reader_base_url_is_valid(baseurl)) {
@@ -109,7 +114,7 @@ tlog_play_create_es_json_reader(struct tlog_errs **perrs,
     query = json_object_get_string(obj);
 
     /* Create the reader */
-    grc = tlog_es_json_reader_create(&reader, baseurl, query, 10);
+    grc = tlog_es_json_reader_create(&reader, baseurl, query, 10, verbose);
     if (grc != TLOG_RC_OK) {
         TLOG_ERRS_RAISECS(grc, "Failed creating the Elasticsearch reader");
     }
--- a/m4/tlog/play_conf_schema.m4
+++ b/m4/tlog/play_conf_schema.m4
@@ -92,6 +92,12 @@ M4_PARAM(`/es', `query', `file-',
          `STRING is the ', `The ',
          `M4_LINES(`query string to send to Elasticsearch')')m4_dnl
 m4_dnl
+M4_PARAM(`/es', `verbose', `file-',
+         `M4_TYPE_BOOL(false)', true,
+         `', `', `Enable verbose output on Elasticsearch HTTP client',
+         `If specified, ', `If true, ',
+         `M4_LINES(`enable verbose output on Elasticsearch HTTP client.')')m4_dnl
+m4_dnl
 m4_dnl
 m4_dnl
 m4_ifelse(M4_JOURNAL_ENABLED(), `1', `m4_dnl
--- a/include/tlog/es_json_reader.h
+++ b/include/tlog/es_json_reader.h
@@ -78,13 +78,14 @@ static inline tlog_grc
 tlog_es_json_reader_create(struct tlog_json_reader **preader,
                            const char *base_url,
                            const char *query,
-                           size_t size)
+                           size_t size,
+                           bool verbose)
 {
     assert(preader != NULL);
     assert(tlog_es_json_reader_base_url_is_valid(base_url));
     assert(query != NULL);
     return tlog_json_reader_create(preader, &tlog_es_json_reader_type,
-                                   base_url, query, size);
+                                   base_url, query, size, verbose);
 }
 
 #endif /* _TLOG_ES_JSON_READER_H */


lib/tlog/es_json_reader.c
```diff
@@ -374,6 +374,10 @@ tlog_es_json_reader_refill_array(struct tlog_es_json_reader *es_json_reader)
     if (getenv("TLOG_CURL_USERPWD"))
         curl_easy_setopt(es_json_reader->curl, CURLOPT_USERPWD, getenv("TLOG_CURL_USERPWD"));
 
+    /* If $TLOG_CURL_VERBOSE is set enable verbose logging on the curl client. */
+    if (getenv("TLOG_CURL_VERBOSE"))
+        curl_easy_setopt(es_json_reader->curl, CURLOPT_VERBOSE, 1);
```
justin-stephenson: This looks okay, but this might be better as an elastic search reader option that can be set in the configuration file, such as **--es-verbose** or **--es-curl-verbose**

Similar to:

~~~
   Elasticsearch reader options
       --es-baseurl=STRING
              Elasticsearch URL without query or fragment parts.

              STRING  is the base URL to request Elasticsearch through. Should not contain query (?...) or fragment
              (#...) parts.

       --es-query=STRING
              Elasticsearch query.

              STRING is the query string to send to Elasticsearch
~~~

lib/tlog/es_json_reader.c
```diff
@@ -364,6 +364,16 @@ tlog_es_json_reader_refill_array(struct tlog_es_json_reader *es_json_reader)
         goto cleanup;
     }
 
+    /* Allow kerberos (negotiate) authentication */
+    curl_easy_setopt(es_json_reader->curl, CURLOPT_HTTPAUTH, CURLAUTH_DIGEST | CURLAUTH_BASIC | CURLAUTH_GSSNEGOTIATE);
+
+    /* Read credentials from .netrc if present */
+    curl_easy_setopt(es_json_reader->curl, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
```
justin-stephenson: We should check the return code of each curl library call
~~~
    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_WRITEFUNCTION,
                          tlog_es_json_reader_write_func);
    if (rc != CURLE_OK) {
        grc = TLOG_GRC_FROM(curl, rc);
        goto error;
    }
~~~

lib/tlog/es_json_reader.c
```diff
@@ -380,12 +381,19 @@ tlog_es_json_reader_refill_array(struct tlog_es_json_reader *es_json_reader)
...
+
     if (rc != CURLE_OK) {
         if (rc == CURLE_WRITE_ERROR && data.rc != json_tokener_success) {
             grc = TLOG_GRC_FROM(json, data.rc);
         } else {
             grc = TLOG_GRC_FROM(curl, rc);
         }
+        if (http_status > 0 && http_status != 200) {
+            fprintf(stderr, "Failed HTTP request to %s (%ld response)", url, http_status);
+        }
```
justin-stephenson: I would prefer to provide a libcurl option like `curl -i` to print the HTTP status response, then let the Administrator decide how to parse that information.

ajf8: OK, I removed this bit. It was kind of useful because tlog-play was trying to parse the data, whatever the response, so you got a "write error" back from curl which is kind of confusing. I set FAILONERROR, which means it now tells you the server responded with an error.

It doesn't tell you what the error is, but the --es-verbose option now provides what is needed.

lib/tlog/es_json_reader.c
```diff
@@ -364,6 +364,16 @@ tlog_es_json_reader_refill_array(struct tlog_es_json_reader *es_json_reader)
...
 
+    /* Allow kerberos (negotiate) authentication */
+    curl_easy_setopt(es_json_reader->curl, CURLOPT_HTTPAUTH, CURLAUTH_DIGEST | CURLAUTH_BASIC | CURLAUTH_GSSNEGOTIATE);
+
+    /* Read credentials from .netrc if present */
+    curl_easy_setopt(es_json_reader->curl, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
+
+    /* Take credentials from TLOG_CURL_USERPWD environment variable if present */
+    if (getenv("TLOG_CURL_USERPWD"))
+        curl_easy_setopt(es_json_reader->curl, CURLOPT_USERPWD, getenv("TLOG_CURL_USERPWD"));
```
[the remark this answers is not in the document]
ajf8: In our case we want to have a wrapper script around tlog-play which takes credentials from a prompt, or maybe from our hashicorp vault.

We don't like putting credentials in commands, they get logged.

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +232,46 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
...
+    /* On >= 400 responses fail immediately, before trying to parse JSON */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_FAILONERROR, 1);
+    if (rc != CURLE_OK) {
+        grc = TLOG_GRC_FROM(curl, rc);
+        goto error;
+    }
+
+    /* Take credentials from TLOG_CURL_USERPWD environment variable if present */
+    if (getenv("TLOG_CURL_USERPWD")) {
+        curl_easy_setopt(es_json_reader->curl, CURLOPT_USERPWD, getenv("TLOG_CURL_USERPWD"));
```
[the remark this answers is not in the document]
spbnick: How about making this setting a full-blown configuration option of the Elasticsearch reader, and adding support for reading configuration from `TLOG_PLAY_CONF_TEXT`, similar to `TLOG_REC_CONF_TEXT`?

Then you would be able to specify it on the command line, in the configuration file, or via the environment with e.g. `TLOG_PLAY_CONF_TEXT='{"es": {"userpwd": "<USER>:<PASSWORD>"}'`.

That is, of course, much more involved, but then we wouldn't open the gates to more arbitrary environment variables sprinkled over the place, which would need separate documentation and wouldn't participate with configuration overrides.

m4/tlog/play_conf_schema.m4
```diff
@@ -92,6 +92,12 @@ M4_PARAM(`/es', `query', `file-',
          `STRING is the ', `The ',
          `M4_LINES(`query string to send to Elasticsearch')')m4_dnl
 m4_dnl
+M4_PARAM(`/es', `verbose', `opts-',
+         `M4_TYPE_BOOL(false)', true,
+         `', `', `Enable verbose output on ElasticSearch HTTP client',
```
spbnick, line 97: Even though I personally like "ElasticSearch" better, it's spelled "Elasticsearch" (with lowercase 's').

m4/tlog/play_conf_schema.m4
```diff
@@ -92,6 +92,12 @@ M4_PARAM(`/es', `query', `file-',
...
+         `If specified, ', `If true, ',
+         `M4_LINES(`enable verbose output on ElasticSearch HTTP client.')')m4_dnl
```
spbnick, line 99: Here as well.

lib/tlog/play.c
```diff
@@ -95,6 +96,11 @@ tlog_play_create_es_json_reader(struct tlog_errs **perrs,
     }
     baseurl = json_object_get_string(obj);
 
+    /* Check for the verbose flag */
+    if (json_object_object_get_ex(conf, "verbose", &obj)) {
+        verbose = json_object_get_boolean(obj);
+    }
+
```
spbnick, line 103: This should rather be something like this:
```c
    verbose = json_object_object_get_ex(conf, "verbose", &obj) &&
              json_object_get_boolean(obj);
```
Otherwise `verbose` could end up uninitialized, if the option is missing.

lib/tlog/es_json_reader.c
```diff
@@ -48,6 +48,7 @@ struct tlog_es_json_reader {
     size_t                      idx;        /**< Index of the message to be
                                                  read next */
     size_t                      last_id;    /**< Last received message ID */
+    bool                        verbose;    /**< Enable HTTP debug */
```
spbnick, line 51: We don't seem to need this beyond initialization, so we don't need to store it. Please remove.

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +232,46 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
...
+        curl_easy_setopt(es_json_reader->curl, CURLOPT_USERPWD, getenv("TLOG_CURL_USERPWD"));
+        if (rc != CURLE_OK) {
+            grc = TLOG_GRC_FROM(curl, rc);
+            goto error;
+        }
+    }
+
+    /* Enable verbose logging on the curl client. */
+    es_json_reader->verbose = verbose;
+    if (verbose) {
```
spbnick, line 267: Can you just always assign the `verbose` value to `CURLOPT_VERBOSE`, and drop this `if`?
E.g.:
```c
    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_VERBOSE, verbose);
```

ajf8: Thank you for reviewing

lib/tlog/es_json_reader.c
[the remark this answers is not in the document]
ajf8: I see your point, but I think credentials are a bit of a special case, and being able to pass them in environment variables is a fairly common pattern. Everything else can be a command line option.

m4/tlog/play_conf_schema.m4
```diff
@@ -92,6 +92,12 @@ M4_PARAM(`/es', `query', `file-',
          `STRING is the ', `The ',
          `M4_LINES(`query string to send to Elasticsearch')')m4_dnl
 m4_dnl
+M4_PARAM(`/es', `verbose', `opts-',
```
justin-stephenson, line 95: Please change the `opts-` here to `file-` to allow this option to be configured in the configuration file.

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +231,43 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
         goto error;
     }
 
+    /* Allow kerberos (negotiate) authentication */
```
justin-stephenson, line 234: Can you please verify this works with kerberos authentication on your side?

ajf8: Yes, I tested this with one of our kerberos enabled ES instances. I'm kind of surprised nobody has requested kerberos support and wonder how others control access to tlog indexes :)

ajf8:
Thanks Justin. I've added checks for those return codes where I think they make sense, though I think you'd have a hard time finding a build of curl which doesn't support them ;)

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +231,45 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
...
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
+    if (rc != CURLE_OK && rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+        grc = TLOG_GRC_FROM(curl, rc);
+        goto error;
```
justin-stephenson, line 238: Minor style change request: Could you do the following instead?
~~~
   rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
    if (rc != CURLE_OK)
        if (rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
            grc = TLOG_GRC_FROM(curl, rc);
            goto error;
   }
~~~

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +231,45 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
...
+    }
+
+    /* Read credentials from .netrc if present */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
+    if (rc != CURLE_OK && rc != CURLE_UNKNOWN_OPTION && rc != CURLE_NOT_BUILT_IN) {
+        grc = TLOG_GRC_FROM(curl, rc);
+        goto error;
+    }
```
justin-stephenson, line 246: Same as commented above.

lib/tlog/es_json_reader.c
```diff
@@ -230,6 +231,45 @@ tlog_es_json_reader_init(struct tlog_json_reader *reader, va_list ap)
...
+
+    /* On >= 400 responses fail immediately, before trying to parse JSON */
+    rc = curl_easy_setopt(es_json_reader->curl, CURLOPT_FAILONERROR, 1);
+    if (rc != CURLE_OK) {
+        grc = TLOG_GRC_FROM(curl, rc);
+        goto error;
+    }
```
justin-stephenson, line 253: Same as commented above.

lib/tlog/es_json_reader.c
justin-stephenson: Was this resolved? I still see 
~~~
    /* Configure verbose logging on the curl client. */
    if (verbose) {
~~~

justin-stephenson:
Thank you @ajf8! Should be good to go after the latest minor changes I commented in-line.

ajf8:
Sure, it's done.

justin-stephenson:
Thanks for the contribution @ajf8 !

Issue #280, opened 2020-03-09
tlog-rec-session does not create an utmp entry, leading the “who am i” command to fail


casantos:
tlog-rec-session allocates a new pseudo-terminal for the recorded user session. Since the _who_ command searches for an entry corresponding the current terminal in the `run/utmp` file, that search always fails. For example, in my test system I have user **newuser**:
```
# getent passwd user newuser
newuser:x:1001:1001::/home/newuser:/usr/bin/tlog-rec-session
```
So a **newuser** session becomes
```
[root@rhel-8-1 tmp]# tty
/dev/pts/0
[root@rhel-8-1 ~]# su - newuser
Locale charset is ANSI_X3.4-1968 (ASCII)
Assuming locale environment is lost and charset is UTF-8

ATTENTION! Your session is being recorded!

[newuser@rhel-8-1 ~]$ id
uid=1001(newuser) gid=1001(newuser) groups=1001(newuser) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[newuser@rhel-8-1 ~]$ tty
/dev/pts/1
[newuser@rhel-8-1 ~]$ ps aufwww|head -n 6
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root      8947  0.0  0.2  26712  5224 pts/0    Ss   17:57   0:00 -bash
root      9356  0.0  0.5 187336 10648 pts/0    S+   18:08   0:00  \_ su - newuser
tlog      9357  0.0  0.4 121780  7488 pts/0    S+   18:08   0:00      \_ -tlog-rec-session
newuser   9358  0.0  0.2  24564  5208 pts/1    Ss   18:08   0:00          \_ -bash
newuser   9436  0.0  0.2  57180  3964 pts/1    R+   18:11   0:00              \_ ps aufwww
[newuser@rhel-8-1 ~]$ who am i
[newuser@rhel-8-1 ~]$ 
```
The solution requires modifying `tlog-rec-session` to add an entry in the utmp file so the heuristics used by the _who_ command succeeds.

PR #285, opened 2020-03-27
[RFC] fix issue #280

casantos:
Add code to tlog_tap_setup and tlog_tap_teardown to add/invalidade utmp
entries for the shell subprocess.

At the moment this requires write permission to /run/utmp, so the user
must belong to the "utmp" group.

Fixes:
  - https://github.com/Scribery/tlog/issues/280
  - https://bugzilla.redhat.com/show_bug.cgi?id=1811794

Signed-off-by: Carlos Santos <casantos@redhat.com>

### DIFF
--- a/include/tlog/tap.h
+++ b/include/tlog/tap.h
@@ -32,18 +32,25 @@
 #include <termios.h>
 #include <unistd.h>
 #include <time.h>
+#include <utmpx.h>
+
+#define TLOG_TAP_UTMP_LINESIZE (sizeof(((struct utmpx *)0)->ut_line))
+#define TLOG_TAP_DEVPATH "/dev/"
+#define TLOG_TAP_PTSPATH TLOG_TAP_DEVPATH "pts/"
+#define TLOG_TAP_DEVSIZE (sizeof(TLOG_TAP_DEVPATH) - 1)
 
 /** I/O tap state */
 struct tlog_tap {
-    pid_t               pid;            /**< Shell PID */
-    struct tlog_source *source;         /**< TTY data source */
-    struct tlog_sink   *sink;           /**< TTY data sink */
-    int                 in_fd;          /**< Shell input FD */
-    int                 out_fd;         /**< Shell output FD */
-    int                 tty_fd;         /**< Controlling terminal FD, or -1 */
-    struct termios      termios_orig;   /**< Original terminal attributes */
-    bool                termios_set;    /**< True if terminal attributes were
-                                             changed from the original */
+    pid_t               pid;                    /**< Shell PID */
+    struct tlog_source *source;                 /**< TTY data source */
+    struct tlog_sink   *sink;                   /**< TTY data sink */
+    struct utmpx       *ut;                     /**< utmp entry */
+    int                 in_fd;                  /**< Shell input FD */
+    int                 out_fd;                 /**< Shell output FD */
+    int                 tty_fd;                 /**< Controlling terminal FD, or -1 */
+    struct termios      termios_orig;           /**< Original terminal attributes */
+    bool                termios_set;            /**< True if terminal attributes were
+                                                     changed from the original */
 };
 
 /** A void I/O tap state initializer */
@@ -51,6 +58,7 @@ struct tlog_tap {
     (struct tlog_tap) {      \
         .source = NULL, \
         .sink   = NULL, \
+        .ut     = NULL, \
         .in_fd  = -1,   \
         .out_fd = -1,   \
         .tty_fd = -1,   \
@@ -59,17 +67,18 @@ struct tlog_tap {
 /**
  * Setup I/O tap.
  *
- * @param perrs     Location for the error stack. Can be NULL.
- * @param ptap      Location for the tap state.
- * @param euid      The effective UID the program was started with.
- * @param egid      The effective GID the program was started with.
- * @param opts      Execution options: a bitmask of TLOG_EXEC_OPT_* bits.
- * @param path      Path to the recorded program to execute.
- * @param argv      ARGV array for the recorded program.
- * @param in_fd     Stdin to connect to, or -1 if none.
- * @param out_fd    Stdout to connect to, or -1 if none.
- * @param err_fd    Stderr to connect to, or -1 if none.
- * @param clock_id  Clock to use for timestamps.
+ * @param perrs        Location for the error stack. Can be NULL.
+ * @param ptap         Location for the tap state.
+ * @param euid         The effective UID the program was started with.
+ * @param egid         The effective GID the program was started with.
+ * @param opts         Execution options: a bitmask of TLOG_EXEC_OPT_* bits.
+ * @param path         Path to the recorded program to execute.
+ * @param argv         ARGV array for the recorded program.
+ * @param in_fd        Stdin to connect to, or -1 if none.
+ * @param out_fd       Stdout to connect to, or -1 if none.
+ * @param err_fd       Stderr to connect to, or -1 if none.
+ * @param clock_id     Clock to use for timestamps.
+ * @param update_utmp  Attempt updating utmp if true.
  *
  * @return Global return code.
  */
@@ -79,7 +88,8 @@ extern tlog_grc tlog_tap_setup(struct tlog_errs **perrs,
                                unsigned int opts,
                                const char *path, char **argv,
                                int in_fd, int out_fd, int err_fd,
-                               clockid_t clock_id);
+                               clockid_t clock_id,
+                               bool update_utmp);
 
 /**
  * Teardown an I/O tap state.
--- a/lib/tlog/tap.c
+++ b/lib/tlog/tap.c
@@ -32,6 +32,133 @@
 #include <time.h>
 #include <semaphore.h>
 #include <assert.h>
+#include <limits.h>
+#include <ctype.h>
+#include <string.h>
+#include <sys/types.h>
+#include <pwd.h>
+
+/**
+ * Add an utmp entry for this session.
+ *
+ * @param tap   tlog_tap structure address.
+ * @param path  The filename of the slave pty.
+ * @param perrs Location for the error stack
+ * @param euid  The EUID to use while updating utmp
+ * @param egid  The EGID to use while updating utmp
+ *
+ * @return Global return code.
+ */
+static tlog_grc
+tlog_tap_add_utmp_entry(struct tlog_tap *tap, char *path,
+                        struct tlog_errs **perrs,
+                        uid_t euid, gid_t egid)
+{
+    tlog_grc grc;
+    int ofs;
+    struct timeval tv;
+    char *ses_line;
+    struct utmpx ut;
+    struct utmpx *ut_ptr;
+
+    if ((ses_line = ttyname(tap->tty_fd)) == NULL) {
+        grc = TLOG_GRC_ERRNO;
+        TLOG_ERRS_RAISECS(grc, "Failed retrieving terminal name");
+    }
+
+    if (gettimeofday(&tv, NULL) < 0) {
+        grc = TLOG_RC_FAILURE;
+        TLOG_ERRS_RAISES("Failure obtaining the current time");
+    }
+
+    if ((tap->ut = calloc(sizeof(*tap->ut), 1)) == NULL) {
+        grc = TLOG_GRC_FROM(errno, ENOMEM);
+        goto cleanup;
+    }
+
+    /* Set basic utmp entry data */
+    memset(&ut, 0, sizeof(struct utmpx));
+    strncpy(ut.ut_user, getpwuid(getuid())->pw_name, sizeof(ut.ut_user));
+    ut.ut_type = USER_PROCESS;
+    ut.ut_pid = tap->pid;
+
+    /* Strip the leading "/dev/" to get the line name */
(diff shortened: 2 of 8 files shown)


lib/tlog/tap.c
```diff
@@ -206,21 +212,70 @@ tlog_tap_setup(struct tlog_errs **perrs,
         }
 
         /* Fork a child connected via a PTY */
-        tap.pid = forkpty(&master_fd, NULL, &tap.termios_orig, &winsize);
+        tap.pid = forkpty(&master_fd, tty_path, &tap.termios_orig, &winsize);
```
justin-stephenson, line 215: Can we follow the warning suggestion mentioned here instead?

https://www.gnu.org/software/libc/manual/html_node/Pseudo_002dTerminal-Pairs.html#Pseudo_002dTerminal-Pairs

lib/tlog/tap.c
```diff
@@ -206,21 +212,70 @@ tlog_tap_setup(struct tlog_errs **perrs,
...
             tap.in_fd = master_fd;
             tap.out_fd = dup(master_fd);
             if (tap.out_fd < 0) {
                 grc = TLOG_GRC_ERRNO;
                 TLOG_ERRS_RAISECS(grc, "Failed duplicating PTY master FD");
             }
+
+            memset(&tap.ut, 0, sizeof(tap.ut));
+
+            ses_line = ttyname(STDIN_FILENO);
```
justin-stephenson, line 238: Please use **tap.tty_fd** instead of **STDIN_FILENO**

include/tlog/tap.h
```diff
@@ -32,18 +32,24 @@
 #include <termios.h>
 #include <unistd.h>
 #include <time.h>
+#include <utmpx.h>
+
+#define TAP_LINESIZE (sizeof(((struct utmpx *)0)->ut_line))
+#define TAP_DEVPATH "/dev/"
+#define TAP_DEVSIZE (sizeof(TAP_DEVPATH) - 1)
```
justin-stephenson, line 39: To follow convention, can we use names prefixed with TLOG_ such as:

~~~
   TLOG_TAP_UTMP_LINESIZE
   TLOG_TAP_DEVPATH
   TLOG_TAP_DEVSIZE
~~~

lib/tlog/tap.c
```diff
@@ -206,21 +212,70 @@ tlog_tap_setup(struct tlog_errs **perrs,
...
+                strcpy(tap.ut.ut_user, getpwuid(getuid())->pw_name);
+                if (ut) {
+                    strcpy(tap.ut.ut_host, ut->ut_host);
+                    tap.ut.ut_session = ut->ut_session;
+                    memcpy(tap.ut.ut_addr_v6, ut->ut_addr_v6, sizeof(tap.ut.ut_addr_v6));
+                }
+
+                res = pututxline(&tap.ut);
+                endutxent();
+            }
```
justin-stephenson, line 278: Can you please add all the code above into its own function? Something like `tlog_tap_utmp_entry_add`

lib/tlog/tap.c
```diff
@@ -372,6 +428,17 @@ tlog_tap_teardown(struct tlog_errs **perrs,
             return grc;
         }
         tap->pid = 0;
+        if (tap->ut.ut_type == USER_PROCESS) {
```
justin-stephenson, line 431: Do we need to make sure tap->ut is non-NULL here?

lib/tlog/tap.c
casantos: The path can't be larger than PATH_MAX, so I will use it as the size of tty_path.

lib/tlog/tap.c
casantos: No, because tap->ut was not a pointer, but I'm changing this.

lib/tlog/tap.c
casantos: adding a static “tlog_tap_add_utmp_entry” function

casantos: I'm also adding a corresponding “tlog_tap_remove_utmp_entry”

casantos:
I updated the code with improved error handling.

lib/tlog/tap.c
```diff
@@ -32,6 +32,117 @@
...
+    endutxent();
+
+    if (ut_ptr != NULL) {
+        memcpy(tap->ut, &ut, sizeof(ut));
+        return 0;
+    }
+
+    free(tap->ut);
+    tap->ut = NULL;
+    return -1;
```
justin-stephenson, line 111: Looking at the source of other login-related applications, we don't want to fail if pututxline() returns NULL - we can ignore failures.

lib/tlog/tap.c
```diff
@@ -32,6 +32,117 @@
...
+        return -1;
+
+    if (gettimeofday(&tv, NULL) < 0)
+        return -1;
+
+    if ((tap->ut = calloc(sizeof(*tap->ut), 1)) == NULL)
+        return -1;
+
+    /* Set basic utmp entry data */
+    strncpy(ut.ut_user, getpwuid(getuid())->pw_name, sizeof(ut.ut_user));
```
justin-stephenson, line 68:
~~~
In file included from /usr/include/string.h:494,                                                                                                                                                                                              
                 from tap.c:37:                                                                                                                                                                                                               
In function 'strncpy',                                                                                                                                                                                                                        
    inlined from 'tlog_tap_add_utmp_entry' at tap.c:68:5,                                                                                                                                                                                     
    inlined from 'tlog_tap_setup' at tap.c:336:17:                                                                                                                                                                                            
/usr/include/bits/string_fortified.h:106:10: warning: '__builtin_strncpy' specified bound 32 equals destination size [-Wstringop-truncation]                                                                                                  
  106 |   return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));                                                                                                                                                              
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~    
~~~

justin-stephenson:
I think it also makes sense to have a build time configure option to disable UTMP, what do you think?

lib/tlog/tap.c
```diff
@@ -32,6 +32,117 @@
...
+     * data to the new entry */
+    ofs = strncmp(ses_line, TLOG_TAP_DEVPATH, TLOG_TAP_DEVSIZE) == 0 ? TLOG_TAP_DEVSIZE : 0;
+    ses_line += ofs;
+    while ((ut_ptr = getutxent()) != NULL &&
+            (ut_ptr->ut_type != USER_PROCESS || strcmp(ut_ptr->ut_line, ses_line)))
+        ;
+    if (ut_ptr) {
+        memcpy(ut.ut_host, ut_ptr->ut_host, sizeof(ut_ptr->ut_host));
+        ut.ut_session = ut_ptr->ut_session;
+        memcpy(ut.ut_addr_v6, ut_ptr->ut_addr_v6, sizeof(ut_ptr->ut_addr_v6));
```
justin-stephenson, line 96: I was wondering why we need this, and also how the `ut_host`, `ut_session`, and `ut_addr_v6` information will be retrieved from a previous entry if we are never originally writing these fields to the utmp record.

That said I am seeing this compiler error when building with `--enable-debug`

~~~
tap.c: In function 'tlog_tap_add_utmp_entry':
tap.c:91:49: error: 'strcmp' argument 1 declared attribute 'nonstring' [-Werror=stringop-overflow=]
91 |             (ut_ptr->ut_type != USER_PROCESS || strcmp(ut_ptr->ut_line, ses_line)))                                                                                                                                                         
In file included from /usr/include/utmpx.h:31,
                 from /home/jstephen/rpmbuild/BUILD/tlog-9/include/tlog/tap.h:35,
                 from tap.c:24:

lib/tlog/tap.c
```diff
@@ -32,6 +32,117 @@
...
+        memcpy(&ut, tap->ut, sizeof(ut));
+        free(tap->ut);
+        tap->ut = NULL;
+        ut.ut_type = DEAD_PROCESS;
+        memset(&ut.ut_user, 0, sizeof(ut.ut_user));
+        memset(ut.ut_host, 0, sizeof(ut.ut_host));
+        ut.ut_exit.e_termination = WIFSIGNALED(wstatus) ? WTERMSIG(wstatus) : 0;
+        ut.ut_exit.e_exit = WEXITSTATUS(wstatus);
+        ut.ut_tv.tv_sec = 0;
+        ut.ut_tv.tv_usec = 0;
```
justin-stephenson, line 138: Shouldn't this be logout time?

justin-stephenson:
Hi Carlos, apologies for the long delay but I now have time to look into this further. I updated my in-line comments. 

One thing I am not sure about is handling a failure in `pututxline()`. with this change should we be adding the tlog user into the utmp group with a rpm script, or do we ignore a failure in `pututxline()` and continue executing **tlog-rec-session** even if the utmp file cannot be written to. Maybe it would be sufficient to print a warning to the administrator, but not block tlog-rec-session entirely in the case the group privileges are not set on the system.

If you would like I can continue working on this PR, or i'm happy to continue as a reviewer. I would also like to add a `--disable-utmp`  configure option, or add build-time support to detect if utmp exists.

lib/tlog/tap.c
casantos: ut_host, ut_session, and ut_addr_v6 are copied from the record created for current session (the one from which tlog-rec-session is executing the user shell. I will replace strcmp by strncmp.

lib/tlog/tap.c
casantos: No. We are disposing the record. I checked in util-linux and login does the same.

casantos:
> Hi Carlos, apologies for the long delay but I now have time to look into this further. I updated my in-line comments.
> 
> One thing I am not sure about is handling a failure in `pututxline()`. with this change should we be adding the tlog user into the utmp group with a rpm script, or do we ignore a failure in `pututxline()` and continue executing **tlog-rec-session** even if the utmp file cannot be written to. Maybe it would be sufficient to print a warning to the administrator, but not block tlog-rec-session entirely in the case the group privileges are not set on the system.

Well, att the moment adding the tlog user to the utmp group is not enough. It is necessary to add the real user to the utmp group.

> If you would like I can continue working on this PR, or i'm happy to continue as a reviewer. I would also like to add a `--disable-utmp` configure option, or add build-time support to detect if utmp exists.

I'd prefer to use a configuration and/or command line option to enable/disable utmp but that's not strong opinion. If you want to contribute to the coding, feel free. :-)

justin-stephenson:
> 
> I'd prefer to use a configuration and/or command line option to enable/disable utmp but that's not strong opinion. If you want to contribute to the coding, feel free. :-)

Thank you, I will add a ./configure and configuration file option for writing to utmp and update this PR branch once done, keeping you as the original author.

casantos:
We need to remove the seteuid(getuid()) and setegid(getgid()) in tlog-rec-session.c to be able to write /run/utmp as the tlog user. That would require to wrap the forkpty call in tap.c in the TLOG_EVAL_WITH_EUID_EGID macro.

justin-stephenson:
The user will need to belong to the "utmp" group as you originally mention, I don't think we want to change tlog-rec-session to be setgid for the 'utmp' group.

justin-stephenson:
@casantos Please review the updates here - only some minor code changes to address a valgrind complaint and preferred use of braces in a 1-line if statement. The configuration option `update-utmp` is disabled by default and only applies to `tlog-rec-session`.

casantos:
@justin-stephenson I fixed the message in your last commit (was signed as “root”) made two additional changes that I believe are useful to fulfill the satisfy the requirements of the original bug and to test the utmp functionality.

casantos:
> Why not keep privileges dropped at startup, and run `pututxline()` with the `TLOG_EVAL_WITH_EUID_EGID` in the case of setgid privileges, if that is the only operation which requires the elevated SUID/SGID privileges ?

The error handling would be much more complex. It would require deep changes in `tlog_tap_add_utmp_entry` and `tlog_tap_remove_utmp_entry`, since `TLOG_EVAL_WITH_EUID_EGID` depends on a local _grc_ variable and on a _cleanup_ label.

justin-stephenson:
@casantos Do you have any problems with https://github.com/Scribery/tlog/pull/285/commits/86ff1cdf342d6520addd80a3a2b1b4fcceecc5c1 ?

I suppose we don't need changes to `tlog_tap_remove_utmp_entry` because in `tlog_tap_setup` the parent process is locking the privileged EGID after forking the recorded child process.

casantos:
> @casantos Do you have any problems with [86ff1cd](https://github.com/Scribery/tlog/commit/86ff1cdf342d6520addd80a3a2b1b4fcceecc5c1) ?
> 
> I suppose we don't need changes to `tlog_tap_remove_utmp_entry` because in `tlog_tap_setup` the parent process is locking the privileged EGID after forking the recorded child process.

That's an excellent solution. I think we should squash it along with the previous commit.

casantos:
Looks good. Is there any run-time test that we should update?

justin-stephenson:
I think this is good to go, thank you very much for your contribution @casantos !

casantos:
OK, but in order to solve BZ1811794 we must also update the spec to install tlog-rec-session SGID utmp, since adding ordinary users to the utmp group would be a security risk.

justin-stephenson:
Okay, I'll submit it into a separate PR.

PR #295, opened 2020-06-08
Use -Og compiler Optimization due to default hardening

justin-stephenson:
I'm not sure if this is necessary, I only ran into this problem when attempting to build a 
rpm with `--enable-debug` enabled. The gcc man page documentation indicates that `-Og` is a suitable debugging optimization.

Using gcc's 0 Optimization (-O0) flag causes warnings when tlog is
build with -D_FORTIFY_SOURCE=2, this causes ./configure --enable-debug
to fail when building a rpm using a spec file:

    warning _FORTIFY_SOURCE requires compiling with optimization

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,7 +59,7 @@ AC_ARG_ENABLE(
 # Output features to preprocessor and compiler
 if test "$tlog_debug_enabled" = "yes"; then
     CPPFLAGS="$CPPFLAGS -UNDEBUG"
-    CFLAGS="$CFLAGS -Wall -Wextra -Werror -g -O0"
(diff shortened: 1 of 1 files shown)


spbnick:
Don't want to stop you here, but here are my two cents.
Why do we build RPMs with `--enable-debug`?
Can we instead just disable `_FORTIFY_SOURCE` for `--enable-debug` by adding `-U_FORTIFY_SOURCE` to the line you're changing?

justin-stephenson:
I was building an rpm to test a debug build on a different RHEL system, it is not a problem I will make the changes myself manually if I need `--enable-debug` when building an rpm.

PR #297, opened 2020-06-24
Install tlog-rec-session with utmp group ownership

justin-stephenson:
Enables support for SGID update-utmp functionality implemented in https://github.com/Scribery/tlog/commit/60cc06f1b073eaa309422e135b4c5ac99f33c937

### DIFF
--- a/.github/workflows/tlog-tests.yml
+++ b/.github/workflows/tlog-tests.yml
@@ -33,6 +33,6 @@ jobs:
     - name: permission setup
       run: sudo chmod g+s /usr/bin/tlog-rec-session
     - name: permission setup
-      run: sudo chown tlog:tlog /var/run/tlog
+      run: sudo chown tlog:utmp /var/run/tlog
     - name: intg tests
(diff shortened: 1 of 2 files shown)


justin-stephenson:
@casantos After thinking about this more, I would prefer to install and run tlog-rec-session with tlog:tlog ownership to run with the least amount of elevated privileges and maintain consistency with previous installs. I propose building with and adding support into tlog for https://github.com/altlinux/libutempter/ which installs with utmp group ownership and will minimize the amount of utmp/wtmp code we need to add into tlog. Unfortunately, I did not think about this earlier and therefore would likely need to revert https://github.com/Scribery/tlog/commit/60cc06f1b073eaa309422e135b4c5ac99f33c937

I can work on the libutempter changes but it may take some additional time, do you have any concerns with this approach?

casantos:
Hum, using libutempter looks like a good approach. Go ahead. I will read the documentation and try to contribute with an alternate solution.

PR #298, opened 2020-06-25
Add libutempter support

justin-stephenson:
Configure tlog at build-time with `--enable-utempter` to enable writing utmp/wtmp entries with libutempter. This reverts https://github.com/Scribery/tlog/commit/60cc06f1b073eaa309422e135b4c5ac99f33c937

The privilege escalation/locking changes are needed because `utempter_remove_record` fails after the EUID privileges were locked due to:

https://github.com/altlinux/libutempter/blob/master/libutempter/utempter.c#L63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ addons:
             - libjson-c-dev
             - libcurl4-gnutls-dev
             - libsystemd-dev
+            - libutempter-dev
             - clang
             - lcov
             - rpm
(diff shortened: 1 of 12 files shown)


justin-stephenson:
@casantos Could you please review? I did not add an `update-utmp` configurable option because it's not clear to me if there is a need for it. It can be added if there is a good use-case for an administrator, or perhaps it can be added later based on user feedback.

casantos:
I just tested it. Works wonderfully!

PR #305, opened 2020-08-06
Add timestamp output facility to tlog-play

jeriksson:
This patch adds the interactive command 'T' to tlog-play. Hitting 'T' during playback enables timestamp output. Whenever playback is paused, a timestamp is output on the console. This timestamp can then be used with -g on the command line to restart playback at the paused location.

### DIFF
--- a/lib/tlog/play.c
+++ b/lib/tlog/play.c
@@ -708,6 +708,7 @@ cleanup:
     return grc;
 }
 
+bool tlog_print_timestamp = false;
 /**
  * Read playback input, interpreting user input.
  *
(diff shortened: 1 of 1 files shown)


jeriksson:
Also, hitting T again disables timestamp output.

jeriksson:
FWIW, the failed check looks unrelated to this patch. The program builds just fine - something seem to be up with the build environment for the rpm.

justin-stephenson:
Thanks for the PR!

Could you please add this timestamp output into the tlog-play man page also? It would be great if this can be added as a general tlog-play option(maybe `--pause-timestamp`) to allow enabling this functionality on initialization with a command-line option, similar to other options like `--paused`.

It would be appreciated if you can move the timespec conversion code in `tlog_play_run` into a function named `tlog_timespec_to_timestr` in **lib/tlog/timestr.c**. Please use macros with helpful names as opposed to magic numbers `60` and `3600` in the code.

Lastly, squash all the commits into a single commit to add this functionality.

PR #328, opened 2021-04-07
add a no-locking option in tlog-rec-session.conf

ajf8:
this disables locking by session id.


trevor-vaughan:
Nice! Shouldn't this be the default since log loss is going to be the default in many (if not most) non-trivial cases?

ajf8:
I'm incined to say yes it should be and happy to change it if the maintainers agree, but I appreciate it could be a major and unexpected change of behavior for some, so I didn't change the defaults, maybe that could happen at a later time and with more consideration.

I'm happy with just having the option and changing tlog-rec-session.conf in our config management. Currently I sed the LOCK_SESS bigflag to be zero before building tlog at present and will continue to do so, but it seems like something that should be an option.

justin-stephenson:
Hi, this seems reasonable. On busy systems with locking disabled are you seeing an excess of duplicate recordings?

trevor-vaughan:
I guess I look at it the other way. Would you rather lose logs or have to ramp it back if things get overwhelming?

ajf8:
Hi,

No, it hasn't resulted in excessive amounts of logs or any duplication for us, just that multiplexed sessions get logged. We have tlog on bastions which are restricted to just SSH and a few other commands, and on hosts which aren't logged into often. It's fine in both of these cases.

I guess there could be duplication on a desktop / development host, if things are running your shell as defined by NSS rather than $SHELL. But we're not interested in tlog on such hosts.

Thanks,

justin-stephenson (changes requested):
Could you please name the option something like `session_locking` which defaults to enabled?

justin-stephenson (changes requested):
Hi, Please use `session_locking` instead of `enable_locking`, The man page description should indicate this enables locking by audit session ID. Please squash into a single commit.

justin-stephenson:
@ajf8 Would you like to write a test for this in https://github.com/Scribery/tlog/blob/master/lib/tlitest/test_tlog_rec_session.py ? You'll need to add this new option https://github.com/Scribery/tlog/blob/3897015d1e00e003b593d26bbae0d8d42b74e863/lib/tlitest/config.py#L185

justin-stephenson (changes requested):
Thanks very much for the tests. Could you please squash you commits. @spoore1 Would you mind reviewing?

ajf8:
Hi @justin-stephenson no problem, but I haven't quite got the tests working yet. I'll try to figure it out and fix them soon.

PR #333, opened 2021-09-20
tlog-play: add jounral namespace support

alexole:
Add possibility to read from a specific journal namespace.


justin-stephenson:
Thank you for the contribution. Overall I don't have issues with the changes, but how do tlog recordings get logged to a non-default journal namespace?

alexole:
Well, I have a bit non-trivial use-case, in which I wrap `tlog-rec-session` into systemd unit which is capable of writing into a separate namespace, so I replace login shell with something like:

```
#!/usr/bin/env bash
LOG_NAMESPACE=audit

if /usr/bin/tty > /dev/null; then
  EXTRA_ARGS="--pty --send-sighup"
else
  EXTRA_ARGS="--pipe -p IgnoreSIGPIPE=no"
fi
sudo /usr/bin/systemd-run -p LogNamespace=$LOG_NAMESPACE --collect --quiet --same-dir \
  -p PAMName=login -p User="$USER" $EXTRA_ARGS /usr/bin/tlog-rec-session.bin "$@"
```

justin-stephenson:
I would like to leave this PR open for a bit to see if any other tlog users may be interested in this support for journal namespaces.

Ideally, we would also add support to log to a non-default journal namespace on the recording side as well, but it looks like there is no API for that yet (only supported with the `LogNamespace` unit file directive).

alexole:
Yes, unfortunately, there is no API for writing into a specific namespace, it looks like it's only possible to use namespaces within systemd units, not sure if this is going to be changed.

PR #334, opened 2021-11-12
Address journal namespace fixes

justin-stephenson:
Fix the hard failure building tlog in distros like RHEL8 due to:
~~~
   configure: error: Package requirements (libsystemd >= 245) were not met:
~~~

Build the -N namespace m4 man page option conditionally, this results in no need to handle catching  -N to tlog-play if namespace support is not built (as it will be an unknown option).


justin-stephenson:
@alexole Please let me know if you are okay with these changes.

PR #341, opened 2022-01-17
CI: Add cppcheck workflow

justin-stephenson:


justin-stephenson:
This cppcheck gh action does not work on pull requests.

PR #365, opened 2024-01-25
configure: correctly handle systemd versions before 245

TvdW:
PKG_CHECK_MODULES macros need to have unique prefixes, or the latter ones are ignored. In configure.ac we were checking for SYSTEMD_JOURNAL twice, once for the presence of systemd in general, then again for namespace support. However, since the first call would already set the variables, the second check would re-use those.

This results in namespace support getting enabled in systemd versions that don't support it, and the build failing.


justin-stephenson:
Thank you for the PR !

PR #370, opened 2024-07-08
test_session_record_pipe_io_stdin: remove sshpass

spoore1:
Removing reliance on sshpass by adding code in test case to generate an ssh key and remove it after testing.

### DIFF
--- a/lib/tlitest/test_tlog_rec_session.py
+++ b/lib/tlitest/test_tlog_rec_session.py
@@ -217,10 +217,14 @@ class TestTlogRecSession:
         """
         text_in_stdio = 'print("hello world")\n'
         text_out = "hello world"
-        p = Popen(['sshpass', '-p', 'Secret123', 'ssh', '-o',
-                   'StrictHostKeyChecking=no',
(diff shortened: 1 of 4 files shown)


lib/tlitest/test_tlog_rec_session.py
```diff
@@ -22,6 +22,25 @@ def utempter_enabled():
...
 
 
+@pytest.fixture
+def gen_ssh_key(request):
+    def _del_ssh_key(user):
+        # Remove ssh key to ensure no tests affected later
+        run(f'sed -i "/stdin_test/d" ~{user}/.ssh/authorized_keys', shell=True)
+        run(f'rm -rf ~/.ssh/id_rsa_{user}*', shell=True)
+
+    def _gen_ssh_key(user):
```
justin-stephenson, line 32: I think it makes more sense to generate the SSH key as part of `src/tlitest/tlitest-setup` then only for `test_session_record_pipe_io_stdin` pass the SSH command-line options which will use the SSH key. This keeps setup code in one place and allows the tests themselves to be simpler.

PR #378, opened 2026-05-06
rec: add watchdog to detect tlog_rec_transfer spin loop after child exit

uncaney:
## Summary

When the wrapped child has exited but buffered packets remain in the position state machine, `tlog_rec_transfer` (`lib/tlog/rec.c`) can spin indefinitely in `tlog_pkt_pos_cmp` / `tlog_sink_write` **without making any syscalls**.

This patch adds a 30-second watchdog: if `tlog_rec_child_exited` is true AND no successful read/write has happened for 30s, the loop breaks with `ETIMEDOUT` and a clear error message. 21 lines added, 0 removed.

## Reproduction (production)

Observed on Ubuntu 24.04 (tlog 14, package `libtlog0`):
- A tlog-recorded interactive shell ran a pipeline like `find / -path "*forgejo*" 2>/dev/null | head -5`
- After `head` exits early and `find` gets `SIGPIPE`, the wrapped `tlog-rec-session` was left burning **99.6% CPU forever**.
- `strace -c -p $PID` over 5 seconds: **zero syscalls**. Pure user-space compute loop.
- `perf top -p $PID`: 99% in `tlog_rec` → `tlog_sink_write` → `tlog_json_chunk_write` → `tlog_pkt_pos_cmp`.
- `/proc/$PID/stat`: `state=R`, `utime/etime ≈ 1.0` (one full core, sustained).
- Multiple instances accumulated 22h+ of CPU waste before being killed by an external watchdog.

## Root cause analysis

In the existing main loop, the only `break` paths after the child exits dont fire when:
- The child has exited
- BUT `tlog_pkt_is_eof(&pkt)` is false (buffered data remains)
- AND `grc == TLOG_RC_OK` (writes succeed but dont advance `tty_pos` past `log_pos`)

Result: infinite loop between two near-done states.

This watchdog prevents the symptom without behavior change for healthy sessions.

## Testing

- Built cleanly on tlog 14 / Ubuntu 24.04.
- Watchdog string verified in compiled lib.
- Smoke test passes on healthy sessions.
- Reproducing the spin in a controlled test rig is non-trivial (requires specific TTY + buffer pattern). External tlog-orphan-killer systemd timer catches runaways in 5 min as defense-in-depth meanwhile.

30-second threshold is conservative; happy to make configurable.


justin-stephenson:
Thank you for the PR @uncaney - Can you please rebase this PR on top of master? I just now pushed a commit to fix the rpmbuild failure in `Integration tests / fedora-build` PRCI

